PR #7, opened 2021-06-01
feat(tx): token metadata

giovanecosta:


serverless.yml
```diff
@@ -13,6 +13,7 @@ provider:
   runtime: python3.8
   profile: serverless
   region: ${env:AWS_DEFAULT_REGION}
+  lambdaHashingVersion: 20201221
```
pedroferreira1, line 16: What's this for?

serverless.yml
```diff
@@ -48,6 +50,16 @@ custom:
     httpPort: ${env:API_PORT}
   pythonRequirements:
     layer: true
+  s3:
+    host: localhost
+    directory: /tmp
```
pedroferreira1, line 55: Is this for testing?

handlers/get_token_meta_handler.py
```diff
@@ -0,0 +1,28 @@
...
+    event: ApiGatewayEvent,
+    _context: LambdaContext,
+    get_token_metadata: Union[GetTokenMetadata, None] = None
+) -> dict:
+
+    get_token_metadata = get_token_metadata or GetTokenMetadata()
+    response = get_token_metadata.get(event.path['hash'])
+
+    if response is None:
+        raise Exception('not_found')
```
pedroferreira1, line 20: I think we should return 404 here, it's better for the client to handle tokens that don't have metadata

pedroferreira1 (approved):

gateways/clients/s3_client.py
```diff
@@ -0,0 +1,31 @@
...
+        :type bucket: str
+        :param file: file path
+        :type file: str
+        :return: the result body
+        :rtype: str
+        """
+        try:
+            response = self.client.get_object(Bucket=bucket, Key=file)
+            return response['Body'].read().decode()
+        except self.client.exceptions.NoSuchKey:
```
luislhl, line 30: We should log when this exception happens

gateways/token_gateway.py
```diff
@@ -0,0 +1,37 @@
...
+
+        if token_metadata_bucket is not None:
+            token_raw_metadata = self.s3_client.load_file(token_metadata_bucket, file)
+            if token_raw_metadata:
+                token_metadata = json.loads(token_raw_metadata)
+                return TokenMetadata.from_dict(token_metadata)
+
+            return None
+
+        raise Exception('No bucket name in config')
```
luislhl, line 37: I think this would improve reading

```suggestion
        if token_metadata_bucket is None:
            raise Exception('No bucket name in config')
            
        token_raw_metadata = self.s3_client.load_file(token_metadata_bucket, file)
        if token_raw_metadata:
            token_metadata = json.loads(token_raw_metadata)
            return TokenMetadata.from_dict(token_metadata)

        return None
```

serverless.yml
luislhl: I saw this in warnings when deploying serverless.
It's a new feature of serverless that we need this line to already enable.

serverless.yml
```diff
@@ -78,10 +90,28 @@ functions:
...
+  get_token_meta_handler:
+    handler: handlers/get_token_meta_handler.handle
+    package:
+      patterns:
+        - 'handlers/get_token_meta_handler.py'
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - s3: ${env:TOKEN_METADATA_BUCKET}
+        event: s3:*
```
luislhl, line 110: We saw during tests that we do not need this.

serverless.yml
```diff
@@ -48,6 +50,16 @@ custom:
...
+  s3:
+    host: localhost
+    directory: /tmp
+
+resources:
+  Resources:
+    NewResource:
+      Type: AWS::S3::Bucket
+      Properties:
+        BucketName: ${env:TOKEN_METADATA_BUCKET}
```
luislhl, line 62: We have decided to create the bucket manually for now, so I think we don't need this, right?

utils/wrappers/aws/api_gateway.py
```diff
@@ -38,6 +38,10 @@ def wrapper(event: dict, context: LambdaContext, *args: Any, **kwargs: Any) -> d
             try:
                 api_gateway_event = ApiGatewayEvent(event, context)
                 result = function_to_call(api_gateway_event, context, *args, **kwargs)  # type: ignore
+
+                result['headers']['Access-Control-Allow-Origin'] = '*'
```
luislhl, line 42: Why are we using `*` here?

Woudn't it be better to specify the website domain? It could be configured as an EnvVar

serverless.yml
giovanecosta: It's a little improvement that will be default in version 3. So it's a good ideia to use now to avoid future problems

https://www.serverless.com/framework/docs/deprecations#default-providerlambdahashingversion

serverless.yml
giovanecosta: Yes! We need this to run serverless offline with s3 emulation. For now it's just nice to emulate and test things manually but when we implement integrations tests, this will be mandatory.

handlers/get_token_meta_handler.py
giovanecosta: raising `'not_found'` string will lead to a 404 response. I think we can improve it with constants

gateways/clients/s3_client.py
giovanecosta: Addes a todo and created an issue for it https://github.com/HathorNetwork/hathor-explorer-service/issues/13

serverless.yml
giovanecosta: Removed

utils/wrappers/aws/api_gateway.py
giovanecosta: I'm not sure if we want to restrict it. I think we can keep it for now and discuss about it soon.

PR #14, opened 2021-06-08
feat(tx): implements token handler

giovanecosta:


gateways/clients/responses/hathor_core/hathor_core_token_response.py
```diff
@@ -0,0 +1,35 @@
...
+    total: Optional[int]
+    transactions_count: Optional[int]
+    melt: Optional[List[dict]]
+    mint: Optional[List[dict]]
+
+    @classmethod
+    def from_dict(cls, dikt: dict) -> 'HathorCoreTokenResponse':
+        return from_dict(data_class=cls, data=dikt)
+
+    def to_domain(self, id: str) -> Token:
```
pedroferreira1, line 23: Didn't understand the name of this method

pedroferreira1 (approved):

gateways/clients/responses/hathor_core/hathor_core_token_response.py
giovanecosta: `Token` is an object of domain (`domain.tx.token`). `HathorCoreTokenResponse` have all attributes based on what is expected from `hathor-core` response. So this method turns the response into a domain object

handlers/get_token_handler.py
```diff
@@ -0,0 +1,29 @@
...
+) -> dict:
+
+    get_token = get_token or GetToken()
+    response = get_token.get(event.path['hash'])
+
+    if response is None:
+        raise Exception('not_found')
+
+    return {
+        "statusCode": 200,
```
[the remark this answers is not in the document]
giovanecosta: The exception above will lead to a `404` 

https://github.com/HathorNetwork/hathor-explorer-service/blob/70f2d3ed269dc44f284b1d9ebbf7d37946be8fe3/utils/wrappers/aws/api_gateway.py#L35

gateways/clients/responses/hathor_core/hathor_core_token_response.py
giovanecosta: I agree that `to_token` is better, so i changed.

handlers/get_token_handler.py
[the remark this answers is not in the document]
jansegre: Is it possible that `response != None` and still `response.status_code != 200`, couldn't an error (404/500/...) on `get_token.get` have some `response` that is not `None`?

Either way I think we should try to improve this in the future, using a wrapper is great, but using bare `Exception`s isn't that great, I think specific exception classes are much better. Or we could also use something like `aiohttp` that has a good set of handy features like those exceptions classes.

PR #15, opened 2021-06-08
Chore: deploy adjustments

luislhl:
Those are some changes we had to made when deploying the service


pedroferreira1 (approved):

serverless.yml
```diff
@@ -107,3 +120,7 @@ functions:
             parameters: 
               paths: 
                 hash: true
+          caching:
```
giovanecosta, line 123: Isn't `cors: true` needed anymore ??

giovanecosta (approved):
Just left a question. All good for me

serverless.yml
luislhl: Good catch, it went missing for some reason.

Besides, the frontend is indeed getting a cors error.
I'll readd this and test again.

PR #16, opened 2021-06-09
test(tx): add missing unit tests for token handler

giovanecosta:


pedroferreira1 (approved):

tests/fixtures/token_factory.py
```diff
@@ -1,21 +1,31 @@
...
+from domain.tx.token import Token, TokenMetadata, TokenNFT, TokenNFTType
 
 fake = Faker()
 
 
-class TokenNFTFactory(factory.Factory):
+def token_symbol(name):
+    first_name = name.split(' ')[0]
+    symbol = re.sub(r'[aeiou\s]', '', first_name).upper()[:5]
+    if len(symbol) < 3:
```
msbrogli, line 14: Why do we need this? I feel we should never change the token name, but I don't have the full picture here. Anyway, add a comment with an example to clarify please.

msbrogli (approved):

tests/fixtures/token_factory.py
giovanecosta: @msbrogli It's just a factory for test purpose. I just tried to make it a little more realistic.

PR #18, opened 2021-06-11
Ci: Create deployment pipeline

luislhl:
This PR adds a workflow on Github Actions to automate the deployment of the service in our backend.

Both our serverless functions and docker containers are deployed.
For serverless, we just use its own deploy command `serverless deploy` with the right configuration.

For the docker containers, we just push a new image to our repository. Check this to learn more about what happens after the image is pushed until deployment is complete: https://github.com/HathorNetwork/ops-tools/issues/20#issuecomment-858750894

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -69,7 +69,7 @@ jobs:
         run: |
           poetry run make check
       - name: Deploy Lambdas Testnet
-        if: github.ref == 'refs/heads/main'
+        if: ${{ job.environment }} == 'testnet'
         run: |
           npx make deploy-lambdas-dev
         env:
@@ -82,7 +82,7 @@ jobs:
           REDIS_DB: 0
           TOKEN_METADATA_BUCKET: token-metadata-testnet
       - name: Deploy Lambdas Mainnet
-        if: startsWith(github.ref, 'refs/tags/v')
+        if: ${{ job.environment }} == 'mainnet'
         run: |
           npx make deploy-lambdas-mainnet
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,110 @@
...
+            ${{ runner.os }}-
+      - name: Install Dependencies
+        run: |
+          pip -q --no-input install poetry
+          make install
+      - name: Linters
+        run: |
+          poetry run make check
+      - name: Deploy Lambdas Testnet
+        if: github.ref == 'refs/heads/main'
```
luislhl, line 72: Using those ifs is something that will probably come up during reviews, so I'll add my thoughts on this already.

There are some other options on how to do this with GIthub Actions, but I still thought this was the best one for now.

The others options would be:

### Create separate jobs for deploy on testnet and mainnet.
In this case, I would have to either repeat all the steps in each job, or try to share between them the steps that are equal.

I didn't thought repeating everything was good.

To share some steps, we would have to have another job with the shared steps, which are the dependencies installation, that produces an artifact in the end, which then we would download in the 'deployment' jobs.

The problem with this is that this artifact would have too much things inside it, like the node_modules and the python dependencies. Besides, `poetry` would need to be reinstalled in each job, or also be added in the artifact.

### Move every env var to the `init` job

This way, we could just have generic deploy steps here, and we would eliminate the `ifs`.

This is not a bad option, but it would probably require a few more things to work. Like possibly a third job, instead of doing it inside `init`, because some secrets depend on the job's `environment` being set, which is what the `init` job is doing.

Besides, the syntax would not be clean.

pedroferreira1: I agree that this is the best solution for now. I was just thinking if there is a way to use the environment 'testnet' or 'mainnet' in the if, give that we already had this conditional check of the main branch or tag before.

pedroferreira1 (approved):

.github/workflows/deploy.yml
luislhl: That's a good point. I think it's possible, will do this and test to confirm

msbrogli (approved):

.github/workflows/deploy.yml
luislhl: Confirmed, I have commited the changes

Issue #23, opened 2021-06-21
CORS 


giovanecosta:
- We need to be able to configure CORS with env vars
- We need to fix CORS headers for error responses

giovanecosta:
https://github.com/HathorNetwork/hathor-explorer-service/pull/26

jansegre:
I'm OK with this task, as long as it solves any actual problem we're having right now.

Otherwise I think we're looking at CORS the wrong way. We have no reason to make it application specific (confirmed by the fact the we're setting it to a static header, it doesn't take any input into account to produce a different value), and as such it should be a deploy setup problem ([API Gateway has setup options for CORS](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html), otherwise if we use a CloudFront in front of the API, it also has CORS setup optons, and if we opt to use a single domain, which I think we should, than we don't need CORS).

Am I missing something?

jansegre:
I'm OK with this task, as long as it solves any actual problem we're having right now.

Otherwise I think we're looking at CORS the wrong way. We have no reason to make it application specific (confirmed by the fact the we're setting it to a static header, it doesn't take any input into account to produce a different value), and as such it should be a deploy setup problem ([API Gateway has setup options for CORS](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html), otherwise if we use a CloudFront in front of the API, it also has CORS setup optons, and if we opt to use a single domain, which I think we should, than we don't need CORS).

Am I missing something? @luislhl @pedroferreira1 @giovanecosta

luislhl:
> I'm OK with this task, as long as it solves any actual problem we're having right now.
> 
> Otherwise I think we're looking at CORS the wrong way. We have no reason to make it application specific (confirmed by the fact the we're setting it to a static header, it doesn't take any input into account to produce a different value), and as such it should be a deploy setup problem ([API Gateway has setup options for CORS](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html), otherwise if we use a CloudFront in front of the API, it also has CORS setup optons, and if we opt to use a single domain, which I think we should, than we don't need CORS).
> 
> Am I missing something? @luislhl @pedroferreira1 @giovanecosta

Yeah, I think your reasoning about CORS not being a application concern makes sense.

PR #26, opened 2021-06-25
feat: add cors configuration by env vars

giovanecosta:
https://github.com/HathorNetwork/hathor-explorer-service/issues/23

### DIFF
--- a/common/configuration.py
+++ b/common/configuration.py
@@ -44,3 +44,5 @@ REDIS_PORT = config('REDIS_PORT', default=None)
 REDIS_DB = config('REDIS_DB', default='0', cast=int)
 
 TOKEN_METADATA_BUCKET = config('TOKEN_METADATA_BUCKET', default=None)
+
+CORS_ALLOWED_ORIGIN = config('CORS_ALLOWED_ORIGIN', default='*')
--- a/common/configuration.py
+++ b/common/configuration.py
@@ -45,4 +45,4 @@ REDIS_DB = config('REDIS_DB', default='0', cast=int)
 
 TOKEN_METADATA_BUCKET = config('TOKEN_METADATA_BUCKET', default=None)
 
-CORS_ALLOWED_ORIGIN = config('CORS_ALLOWED_ORIGIN', default='*')
+CORS_ALLOWED_ORIGIN = config('CORS_ALLOWED_ORIGIN', default=None)
--- a/serverless.yml
+++ b/serverless.yml
@@ -28,6 +28,7 @@ provider:
     REDIS_PORT: ${env:REDIS_PORT}
     REDIS_DB: ${env:REDIS_DB}
     TOKEN_METADATA_BUCKET: ${env:TOKEN_METADATA_BUCKET}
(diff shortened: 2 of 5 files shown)


common/configuration.py
```diff
@@ -44,3 +44,5 @@ def default(cls) -> 'Environment':
 REDIS_DB = config('REDIS_DB', default='0', cast=int)
 
 TOKEN_METADATA_BUCKET = config('TOKEN_METADATA_BUCKET', default=None)
+
+CORS_ALLOWED_ORIGIN = config('CORS_ALLOWED_ORIGIN', default='*')
```
pedroferreira1, line 48: Maybe the default should be `null` or `''` (I am not sure what's the opposite of `*`) so we never deploy  with `*` by accident?

giovanecosta: I think that the default behavior must be always work. It would be very bad if we just make all the service stop working by a simple mistake and i don't think it's too bad to have CORS allowed to everybody for a while.

serverless.yml
```diff
@@ -28,6 +28,7 @@ provider:
     REDIS_PORT: ${env:REDIS_PORT}
     REDIS_DB: ${env:REDIS_DB}
     TOKEN_METADATA_BUCKET: ${env:TOKEN_METADATA_BUCKET}
+    CORS_ALLOWED_ORIGIN: ${env:CORS_ALLOWED_ORIGIN}
```
luislhl, line 31: We have to set this new variable in `.github/workflows/deploy.yml`

common/configuration.py
luislhl: The problem is that if the default is `*` and someone makes a mistake of not setting something different, we will hardly notice the mistake.

And this is a security feature that shouldn't be kept as `*` without notice.

I tend to agree with @pedroferreira1 's suggestion on this one.

luislhl (changes requested):
We should set the env var in the deploy pipeline

common/configuration.py
jansegre: On the other hand `*` could be susceptible to some attacks, and if we accidentally deploy with it everything will work and we won't notice it until we have to deal with it.

Ideally the deploy should fail if it isn't set to a correct value. And also ideally, it should just work when developing.

Also, IMO we should only have one domain (i.e. not split the explorer into `api.explorer...` and `explorer...`). This would make it so we don't have to set up CORS (which is the safest option, attack-wise).

jansegre (approved):
All in all, I'm OK with these changes as is. Mostly because we're already using `*` CORS on the current deploy of the explorer server, and if this solves any immediate problem, then so be it. But I also feel like this isn't the right place for this, besides my opinion that we should only use one domain, I think it would be better if the API domain (if it is separate) is also on CloudFront, like the frontend is, and being on CloudFront it can add the CORS headers we want.

common/configuration.py
giovanecosta: OK, i'm with the team.

serverless.yml
giovanecosta: Done

luislhl (approved):

PR #42, opened 2021-07-07
feat: data aggregator

giovanecosta:


domain/network/network.py
```diff
@@ -0,0 +1,105 @@
+from dataclasses import asdict, dataclass
+from typing import List
+from datetime import datetime, timedelta
+from dacite import from_dict
+
+from domain.network.node import NodeState, Peer, Node
+
+
+def uptime_to_up_since(uptime: float) -> int:
```
msbrogli, line 9: Missing docstring.

domain/network/network.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def add_connected_to(self, node_id: str) -> None:
+        self.connected_to.append(node_id)
+
+    def remove_connected_to(self, node_id: str) -> None:
+        self.connected_to.remove(node_id)
+
+
+@dataclass
+class AggregatedNode:
```
msbrogli, line 60: What's the different between an AggregatePeer and an AggregateNode? Maybe add docstrings to both of them explaining in which case they are used and what for.

domain/network/network.py
```diff
@@ -0,0 +1,105 @@
...
+    def to_dict(self) -> dict:
+        """ Convert a AggregatedPeer instance into dict
+
+        :return: The dict representation of the AggregatedPeer
+        :rtype: dict
+        """
+        return asdict(self)
+
+    def add_connected_to(self, node_id: str) -> None:
+        self.connected_to.append(node_id)
```
msbrogli, line 53: Maybe use a set? This can get duplicated entries.

text/0002-data-aggregator.md
```diff
@@ -0,0 +1,72 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+We could do the aggregation on client side. However, more requests must be made every second, or always return everyhthing.
+
+The disadvantage of this approach is that we must keep polling. 
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Nodes are still manually setted by us. It would be good to discover new http responding nodes automatically as they join the network.
```
msbrogli, line 67: Maybe it should get the list from an S3 location? Or a redis key? So we can build a separate script to update the list. @luislhl is working in a full-node discovery for other services. Maybe you can use the same solution with same flags to indicate which full-nodes will be used by each service.

text/0002-data-aggregator.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+WebSocket. Everytime `Network` have new content and is saved, a handler can be triggered to broadcast new data to connected clients.
```
msbrogli, line 72: I really don't see many advantages using websocket here since the uptime is changing all the time. And the API is so fast that it doesn't matter the hit rate.

text/0002-data-aggregator.md
```diff
@@ -0,0 +1,72 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+Data aggregator will get data collected by `data-collector`, aggregate them into a Network dataclass, compare with saved data and save it if different.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Three new dataclasses will be created: `AggregatedNode`, `AggregatedPeer` and `Network`.
+
+- `AggregatedNode` and `AggregatedPeer` will be a bit smaller version of `Node` and `Peer` respectively, having `up_since` as the main change.
```
msbrogli, line 27: Nice. Add a description of each field and how they are aggregated, please.

msbrogli, line 27: I'm still trying to figure out why we need an `AggregatedNode` (and how it differs from an `AggregatedPeer`). Can you help me here, please?

text/0002-data-aggregator.md
```diff
@@ -0,0 +1,72 @@
...
+We could do the aggregation on client side. However, more requests must be made every second, or always return everyhthing.
+
+The disadvantage of this approach is that we must keep polling. 
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+- Nodes are still manually setted by us. It would be good to discover new http responding nodes automatically as they join the network.
+
+# Future possibilities
```
msbrogli, line 69: I think we should add that we'd like to show the peers in a world map. This is the coolest part of this aggregation in my point of view.

usecases/aggregate_node_data.py
```diff
@@ -0,0 +1,21 @@
...
+        self.node_gateway = node_gateway or NodeGateway()
+
+    def aggregate(self) -> Union[bool, None]:
+        new_network = self.node_gateway.aggregate_network()
+        old_network = self.node_gateway.get_network()
+
+        networks_diff = DeepDiff(new_network, old_network)
+
+        if networks_diff:
+            return self.node_gateway.save_network(new_network)
```
msbrogli, line 19: I'd say we would enter here "all the time". Should we store a hit counter to see if this diff is really necessary?

usecases/aggregate_node_data.py
```diff
@@ -0,0 +1,21 @@
...
+from deepdiff import DeepDiff
+from gateways.node_gateway import NodeGateway
+
+
+class AggregateNodeData:
+
+    def __init__(self, node_gateway: Union[NodeGateway, None] = None) -> None:
+        self.node_gateway = node_gateway or NodeGateway()
+
+    def aggregate(self) -> Union[bool, None]:
```
msbrogli, line 12: Missing docstring. What's the difference from returning a bool or a None?

usecases/aggregate_node_data.py
```diff
@@ -0,0 +1,21 @@
+from typing import Union
+
+from deepdiff import DeepDiff
+from gateways.node_gateway import NodeGateway
+
+
+class AggregateNodeData:
```
msbrogli, line 7: Missing docstring.

msbrogli (changes requested):

msbrogli:
I haven't reviewed all the files. I make a few comments in specific files but the most important comments are in the design.

domain/network/network.py
```diff
@@ -0,0 +1,105 @@
...
+            id=peer.id,
+            app_version=peer.app_version,
+            up_since=uptime_to_up_since(peer.uptime),
+            address=peer.address,
+            state=peer.state,
+            last_message=int(peer.last_message),
+            latest_timestamp=peer.latest_timestamp,
+            sync_timestamp=peer.sync_timestamp,
+            warning_flags=peer.warning_flags,
+            entrypoints=[],  # TODO: add when Peer entrypoints is fixed HathorNetwork/hathor-explorer-service#28
```
pedroferreira1, line 40: I didn't see this fix in the week tasks. I believe the entrypoints should be fixed before we set data aggregator as completed.

giovanecosta: Good point, it's done now: https://github.com/HathorNetwork/hathor-explorer-service/pull/43

usecases/aggregate_node_data.py
giovanecosta: `StrictRedis set` has bool or None as returning type. It's just to match that. We can probably abstract that to only bool. I just didn't think about it.

text/0002-data-aggregator.md
giovanecosta: Today we are using env vars. I like a lot this "full-node discovery" thing.

text/0002-data-aggregator.md
giovanecosta: I'm proposing we use `up_since` now exactly to avoid this issue and optimize lambda usage and server-client communication

text/0002-data-aggregator.md
giovanecosta: I'll add that

text/0002-data-aggregator.md
giovanecosta: Aggregated Nodes and Peers has less properties and use `up_since` instead of `uptime`.

@pedroferreira1 Node is what we can connect and read the status. Peer is what we only know that exists because it is connected to a Node.

usecases/aggregate_node_data.py
giovanecosta: I don't think so, as we introducing `up_since` instead of `uptime` for comparison

text/0002-data-aggregator.md
```diff
@@ -0,0 +1,72 @@
...
+
+Data aggregator will get data collected by `data-collector`, aggregate them into a Network dataclass, compare with saved data and save it if different.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+Three new dataclasses will be created: `AggregatedNode`, `AggregatedPeer` and `Network`.
+
+- `AggregatedNode` and `AggregatedPeer` will be a bit smaller version of `Node` and `Peer` respectively, having `up_since` as the main change.
+  As `uptime` is always changing as time goes by, `up_since` remains the same.
```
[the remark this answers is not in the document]
giovanecosta: Oh, sure, that's why we'll use `up_since` so we can check the diff and the reconnection will be detected naturally.

text/0002-data-aggregator.md
giovanecosta: Added

domain/network/network.py
```diff
@@ -0,0 +1,200 @@
...
+    up_since_date = datetime.now() - timedelta(seconds=uptime)
+    return int(datetime.timestamp(up_since_date))
+
+
+@dataclass
+class AggregatedPeer:
+    """Aggregated data of a Peer retrieved by requesting a Node
+
+    :param id: id of peer
+    :type id: str
```
msbrogli, line 28: I don't think we need the `type` anymore since we are using type annotations in the code. It will probably get outdated in the future.

domain/network/network.py
```diff
@@ -0,0 +1,200 @@
...
+    :param entrypoints: list entrypoints, if any
+    :type entrypoints: List[str]
+
+    :param connected_to: list of ids of nodes that peer is conected to
+    :type connected_to: Set[str]
+
+    """
+    id: str
+    app_version: str
+    up_since: int
```
msbrogli, line 63: Why store `up_since` instead of `uptime`?

tests/unit/usecases/test_collect_nodes_statuses.py
```diff
@@ -20,6 +22,7 @@ def node_gateway(self):
         return MagicMock()
 
     @patch('usecases.collect_nodes_statuses.HATHOR_NODES', ['a', 'b'])
+    @freeze_time(datetime.now())
```
msbrogli, line 25: Why do you need to freeze the time? I mean, you are not using a specific point in time in the tests.

usecases/aggregate_node_data.py
msbrogli: I got it. But only if we collect data twice or more in the same second, right? Otherwise, even the `up_since` will be different. Maybe we should create a comparison method that receives a tolerance parameter in seconds (e.g. `compareNodes(node1, node2, tol=60)`). What do you think?

msbrogli (changes requested):

domain/network/network.py
giovanecosta: `Node` is what we can request, `Peer` is what we only have data through the `Node`. Aggregated versions follows the same concept. I've improved docstring making it clearer

domain/network/network.py
giovanecosta: done. thanks!

usecases/aggregate_node_data.py
giovanecosta: `up_since` is supposed to be fixed for a running node. It's the timestamp when it started to run.
As we calculate it in runtime using `uptime`,  two instances of `AggregatedNode` for the same node can be different for milliseconds. That's why `up_since` is also `int` instead of `float`.
We still have the edge case where this milliseconds difference also change the integer part of the second but i don't think it's a problem.

tests/unit/usecases/test_collect_nodes_statuses.py
giovanecosta: As we calculate `up_since` in runtime, it can change a second for two different instances of the same Node created in two different points of the code. It's an edge case that can bring intermittence to tests. So freeze the time will avoid it.

domain/network/network.py
giovanecosta: `uptime` keep changing every second. `up_since` is the timestamp of when the node started (`now - uptime`) so it wont change for the same running node.

domain/network/network.py
giovanecosta: Should we remove it? I don't have an opinion about it yet. As we're not generating any doc and the IDE is pretty much fine with the type hints, i think it doesn't matter to me, we can remove or keep it.

text/0002-data-aggregator.md
giovanecosta: Done

PR #44, opened 2021-07-14
feat: metadata api

giovanecosta:
[Rendered Design](https://github.com/HathorNetwork/hathor-explorer-service/blob/7a43b8424c0679e198e3ce8f36c4fe0dbaaa4212/text/0003-metadata-api.md)

We'll keep current token meta endpoint for retro compatibility and remove it once the explorer is reaching the new endpoint.

# Before Merge
- [ ] Create a PR removing the old token meta endpoint and things related to them
- [ ] Create new bucket and folders
- [ ] Copy meta jsons to new bucket

# After Merge
- [ ] Delete old bucket


msbrogli (changes requested):

gateways/metadata_gateway.py
```diff
@@ -0,0 +1,70 @@
...
+        return TokenMetadata.from_dict(token_metadata)
+
+    def _get_metadata(self, s3_object_name: str) -> Union[dict, None]:
+        metadata_bucket = self._metadata_bucket()
+
+        raw_metadata = self.s3_client.load_file(metadata_bucket, s3_object_name)
+        if raw_metadata is None:
+            return None
+
+        return json.loads(raw_metadata)
```
pedroferreira1, line 62: If the data in the metadata file is not a json, this will raise `json.decoder.JSONDecodeError`, right? What happens then? I think we should capture this specific exception and handle it properly maybe returning None (that will return a 404)

domain/metadata/transaction_metadata.py
```diff
@@ -0,0 +1,33 @@
+from dataclasses import dataclass
+from typing import Optional
+
+from domain.metadata.metadata import Metadata, MetadataType
+
+
+@dataclass
+class MetaTransaction:
```
pedroferreira1, line 8: If we add any new property in transaction metadata we need to update the explorer service? Imagine we want to mark a tx as banned or as important (don't know why but just imagining possible situations). How do we handle this?

serverless.yml
```diff
@@ -149,3 +149,26 @@ functions:
...
+      - http:
+          path: metadata/{type}/{hash}
+          method: get
+          cors: true
+          request:
+            parameters:
+              paths:
+                type: true
+                hash: true
+          caching:
```
msbrogli, line 170: Who does this caching? How long does it last?

handlers/get_metadata_handler.py
```diff
@@ -0,0 +1,31 @@
...
+
+from usecases.get_metadata import GetMetadata
+from utils.wrappers.aws.api_gateway import ApiGateway, ApiGatewayEvent
+
+
+@ApiGateway()
+def handle(
+    event: ApiGatewayEvent,
+    _context: LambdaContext,
+    get_metadata: Union[GetMetadata, None] = None
```
msbrogli, line 14: Use `Optional[GetMetadata]`.

handlers/get_metadata_handler.py
```diff
@@ -0,0 +1,31 @@
...
+) -> dict:
+
+    get_metadata = get_metadata or GetMetadata()
+    type = event.path['type']
+    hash = event.path['hash']
+    response = get_metadata.get(type, hash)
+
+    if response is None:
+        raise Exception('not_found')
```
msbrogli, line 23: Should we raise an exception in the lambda? Can't it be confused with an unhandled exception? I feel the lambda will return 500 instead of the 404 expected in the design.

gateways/token_gateway.py
```diff
@@ -1,48 +1,19 @@
...
 from gateways.clients.responses.hathor_core.hathor_core_token_response import HathorCoreTokenResponse
-from gateways.clients.s3_client import S3Client
 
 
 class TokenGateway:
     """Gateway for Token
 
-    :param s3_client: Client for s3 manipulation, default to domain S3Client
-    :type s3_client:
+    :param hathor_core_client: Client for make hathor-core requests, default to domain HathorCoreClient
```
msbrogli, line 11: HathorCoreClient?

gateways/metadata_gateway.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def _metadata_bucket(self) -> str:
+        metadata_bucket = METADATA_BUCKET
+
+        if metadata_bucket is None:
```
msbrogli, line 67: Isn't it better to validate all configuration in the constructor? So we can assume everything is right when we get to this point?

gateways/metadata_gateway.py
```diff
@@ -0,0 +1,70 @@
...
+            return None
+
+        return TransactionMetadata.from_dict(transaction_metadata)
+
+    def get_token_metadata(self, id: str) -> Union[TokenMetadata, None]:
+        """Retrieve token metadata from a json file stored in s3
+
+        :param id: token id
+        :type id: str
+        :raises Exception: The name of the bucket used to store the jsons must be on config
```
msbrogli, line 43: I don't feel this error should belong to this method.

gateways/metadata_gateway.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        folder = TYPE_TO_FOLDER[MetadataType.TRANSACTION]
+        transaction_metadata = self._get_metadata(f"{folder}/{id}.json")
+
+        if transaction_metadata is None:
+            return None
+
+        return TransactionMetadata.from_dict(transaction_metadata)
+
+    def get_token_metadata(self, id: str) -> Union[TokenMetadata, None]:
```
msbrogli, line 38: Use `Optional[TokenMetadata]`.

domain/metadata/transaction_metadata.py
```diff
@@ -0,0 +1,33 @@
...
+    id: str
+    context: Optional[str]
+    genesis: Optional[bool]
+
+
+@dataclass
+class TransactionMetadata(Metadata):
+    data: Optional[MetaTransaction] = None
```
msbrogli, line 16: Why are we using two classes? I feel it would be simper to use one class with all fields there. Well, we can group in subclasses for specific cases if there are too many fields.

msbrogli (changes requested):

text/0003-metadata-api.md
```diff
@@ -0,0 +1,61 @@
...
+
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+Every entity that potentialy have some metadata will make a request and only a few will have, so, the massive majority of requests will return `404`. 
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+We can store it in a database and have some tool to manage it like a CMS.
```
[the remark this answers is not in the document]
giovanecosta: Well, we'll have a database to explorer service anyway. I don't think scalability is a problem for metadata. We have other things to be stored that will increase the size much faster. 

Where this tool would run? I like the CMS because we easily have things like roles, history and is non-programmer friendly, people outside the dev team can manage this data for us.

usecases/get_metadata.py
```diff
@@ -0,0 +1,27 @@
+from typing import Union
+
+from gateways.metadata_gateway import MetadataGateway
+
+
+class GetMetadata:
+
+    def __init__(self, metadata_gateway: Union[MetadataGateway, None] = None) -> None:
```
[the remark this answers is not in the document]
giovanecosta: done

usecases/get_metadata.py
```diff
@@ -0,0 +1,27 @@
...
+        self.metadata_gateway = metadata_gateway or MetadataGateway()
+
+    def get(self, type: str, id: str) -> Union[dict, None]:
+        metadata_methods = {
+            'token': self.metadata_gateway.get_token_metadata,
```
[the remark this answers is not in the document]
giovanecosta: The gateway is the same, only the method changes. Each method return a different type of meta.

usecases/get_metadata.py
```diff
@@ -0,0 +1,27 @@
...
+            'transaction': self.metadata_gateway.get_transaction_metadata
+        }
+
+        meta = None
+
+        method = metadata_methods.get(type, None)
```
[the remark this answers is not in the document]
giovanecosta: changed

usecases/get_metadata.py
```diff
@@ -0,0 +1,27 @@
...
+
+        if method:
+            meta = method(id)
+
+        if meta:
+            return meta.to_dict()
```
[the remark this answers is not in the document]
giovanecosta: because the meta structure changed a little

usecases/get_token_metadata.py
```diff
@@ -1,17 +1,17 @@
...
 
-from gateways.token_gateway import TokenGateway
+from gateways.metadata_gateway import MetadataGateway
 
 
 class GetTokenMetadata:
 
-    def __init__(self, token_gateway: Union[TokenGateway, None] = None) -> None:
-        self.token_gateway = token_gateway or TokenGateway()
+    def __init__(self, metadata_gateway: Union[MetadataGateway, None] = None) -> None:
```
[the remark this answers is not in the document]
giovanecosta: done

usecases/get_token_metadata.py
```diff
@@ -1,17 +1,17 @@
 from typing import Union
 
-from gateways.token_gateway import TokenGateway
+from gateways.metadata_gateway import MetadataGateway
 
 
 class GetTokenMetadata:
```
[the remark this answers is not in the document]
giovanecosta: done

.github/workflows/deploy.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           REDIS_HOST: ${{ secrets.REDIS_HOST }}
           REDIS_PORT: 6379
           REDIS_DB: 0
-          TOKEN_METADATA_BUCKET: token-metadata-testnet
+          METADATA_BUCKET: metadata-testnet
```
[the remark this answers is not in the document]
pedroferreira1: It's still missing some
image

domain/metadata/token_metadata.py
```diff
@@ -0,0 +1,76 @@
...
+    :type verified: Optional[bool]
+
+    :param banned: If token is banned or not. None and False are equivalent
+    :type banned: Optional[bool]
+
+    :param reason: Bannishment reason
+    :type reason: Optional[str]
+
+    :param nft: NFT data, if any.
+    :type nft: Optional[:py:class:`domain.metadata.token_metdata.TokenNFT`]
```
pedroferreira1, line 45: I was thinking about my task nowadays. I need to identify a NFT token even if it's not to show the digital asset on the explorer, i.e. the metadata of the token must have a way for me to know that it's a NFT without having this `TokenNFT` attribute set. 

What's the best way to achieve it? Having a boolean attribute `is_nft` and change this parameter to `nft_data`? Or maybe create a `type` that nowadays can be `NFT` or `NORMAL/DEFAULT`?

gateways/metadata_gateway.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        folder = TYPE_TO_FOLDER[MetadataType.TRANSACTION]
+        transaction_metadata = self._get_metadata(f"{folder}/{id}.json")
+
+        if transaction_metadata is None:
+            return None
+
+        return TransactionMetadata.from_dict(transaction_metadata)
+
+    def get_token_metadata(self, id: str) -> Optional[TokenMetadata]:
```
pedroferreira1, line 38: This method and the one above are exactly the same, right? Maybe we can create a generic method that receives `id`, `folder` and `cls` (the class to call `from_dict`).

gateways/metadata_gateway.py
pedroferreira1: All methods need this constant for now, right?

text/0003-metadata-api.md
```diff
@@ -0,0 +1,62 @@
...
+A request made to the API passing `id` and `type` will retrieve the information or `404` if not found. 
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+**Endpoint:**
+
+GET `/metadata/:type/:id` return `200` | `404`
+
+`metadata-handler` return the stored data
```
pedroferreira1, line 33: Maybe we can have a real example with different types of return. One example for a transaction metadata and another one for a token that has NFT data.

pedroferreira1 (changes requested):

domain/metadata/token_metadata.py
giovanecosta: The NFT you are talking about is not to be shown in explorer but still have a file, right? Even if not, we still can use `type` to achieve that. We only have `AUDIO`, `VIDEO` and `IMAGE` today, and only these are shown on explorer. We can add something like `OTHER` or other more specific type and it will just not be shown on explorer.

gateways/metadata_gateway.py
giovanecosta: @pedroferreira1 yes, but i still feel it's not good to check in constructor. We don't have this var in test env, for instance and i don't think we must to. Of course for some tests we have to mock it, but some tests just don't care about it and i think it doesn't make sense to mock a var only for the class to not break.

I know it's just test but i think it's a good indication why maybe it's not so good to check this on constructor.

gateways/metadata_gateway.py
giovanecosta: `_get_metadata` is already generic. It receives an s3 object path. Those two methods just calculate this paths and returns it as the right class. They are very similar because their jsons are stored in the same way. The ideia is each one know how to calculate your own object path, use `_get_metadata` (or maybe another method) to retrieve it and return the right class having the data

text/0003-metadata-api.md
```diff
@@ -0,0 +1,129 @@
...
+# Summary
+[summary]: #summary
+
+Metadata API will provide metadata information about Transactions, Tokens, Blocks and other things on Network.
+It will start with Transactions and Tokens
+
+# Motivation
+[motivation]: #motivation
+
+We need to have some data to be used on Public Explore and Wallets that can't be stored on the blockchain about
```
pedroferreira1, line 16: Typo "Public Explorer"

domain/metadata/token_metadata.py
```diff
@@ -0,0 +1,88 @@
...
+class TokenNFTMedia:
+    """Data of token NFT
+
+    :param type: NFT Type. image or video.
+    :type type: :py:class:`domain.metadata.token_metdata.TokenNFTMediaType`
+
+    :param file: Media file of the NFT
+    :type file: str
+
+    :param loop: If media will play in loop or not
```
pedroferreira1, line 24: Works only for video medias? Or should we have this for audio as well?

We should make explicit here and in the token meta example in the RFC

text/0003-metadata-api.md
```diff
@@ -0,0 +1,129 @@
...
+  "data": {
+    "id": "00003aa356c9493464c657873b115c5e8667adf58cceeb4b37a1cdae0ddc9536",
+    "verified": false,
+    "banned": false,
+    "reason": null,
+    "nft": true,
+    "nft_media": {
+      "type": "IMAGE",
+      "file": "http://curtis-white.com/mean/officer.png",
+      "loop": false
```
pedroferreira1, line 88: Is it required even for image files? Does it work only for videos or audios can play in loop as well?

domain/metadata/token_metadata.py
```diff
@@ -0,0 +1,88 @@
...
+    :param verified: If token is verified or not. None and False are equivalent
+    :type verified: Optional[bool]
+
+    :param banned: If token is banned or not. None and False are equivalent
+    :type banned: Optional[bool]
+
+    :param reason: Bannishment reason
+    :type reason: Optional[str]
+
+    :param nft: If roken is a nft or not. None and False are equivalent
```
pedroferreira1, line 48: Typo "roken"

domain/metadata/token_metadata.py
```diff
@@ -0,0 +1,88 @@
...
+    :type nft: Optional[:py:class:`domain.metadata.token_metdata.TokenNFTMedia`]
```
pedroferreira1, line 49: Type is `Optional[bool]`

domain/metadata/token_metadata.py
```diff
@@ -0,0 +1,88 @@
...
+
+        :return: The new instance
+        :rtype: :py:class:`domain.metadata.token_metdata.TokenMetadata`
+        """
+        data = dikt.get('data', {})
+        if data.get('nft_media'):
+            data['nft_media'] = TokenNFTMedia(
+                TokenNFTMediaType(data['nft_media']['type'].upper()),
+                data['nft_media']['file'],
+                data['nft_media']['loop']
```
pedroferreira1, line 82: What if it's an image NFT? Do I still need to fill `loop: false` in the metadata json? I feel we should make optional for the fields that are not required.

pedroferreira1 (changes requested):

.github/workflows/deploy.yml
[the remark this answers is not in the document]
giovanecosta: I think it covers all, but i'll be more specific

domain/metadata/token_metadata.py
giovanecosta: For audio too. Ok added more explanation

domain/metadata/token_metadata.py
giovanecosta: fixed

domain/metadata/token_metadata.py
giovanecosta: changed

domain/metadata/token_metadata.py
giovanecosta: Surw, changed to `data['nft_media'].get('loop', None)`

text/0003-metadata-api.md
giovanecosta: fixed.

text/0003-metadata-api.md
giovanecosta: It's not required in s3 json files but will be always present on responses.

pedroferreira1: I think the documentation must have the details for each field.

giovanecosta: Added the datails on fields structure explanation

usecases/get_metadata.py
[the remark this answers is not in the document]
msbrogli: Thanks!

serverless.yml
msbrogli: Seems good. Thanks!

tests/unit/domain/metadata/test_token_metadata.py
```diff
@@ -0,0 +1,34 @@
+from domain.metadata.token_metadata import TokenMetadata
+from tests.fixtures.metadata_factory import MetaTokenFactory, TokenMetadataFactory, TokenNFTFactory
+
+
+class TestTokenMetadata:
```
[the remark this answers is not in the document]
msbrogli: But not using it forbids us from using many testing tools. For instance, when the first test fails, it stops running. We can't select and run one specific test. I can't see why we wouldn't use the tools designed to do this work.

text/0003-metadata-api.md
```diff
@@ -0,0 +1,130 @@
...
+{
+  "id": string,                         // hash id of token or transaction 
+  "type": enum(TOKEN|TRANSACTION),      // type of metadata
+  "data": (TokenMeta|TransactionMeta)   // metadata depending on type
+}
+```
+
+```
+TokenMeta {                             // if type is token
+  id: string,                           // hash id of token or transaction
```
pedroferreira1, line 53: Suggestion: 'hash id of token or transaction ' -> 'token uid'

text/0003-metadata-api.md
```diff
@@ -0,0 +1,130 @@
...
+```
+
+```
+TokenMeta {                             // if type is token
+  id: string,                           // hash id of token or transaction 
+  verified: boolean,                    // if token is verified or not
+  banned: boolean,                      // if token is banned
+  reason: string,                       // reason of ban
+  nft: boolean,                         // if token is a NFT
+  nft_media: {                          // media data of NFT
```
pedroferreira1, line 58: Suggestion: 'media data of NFT (optional)'

text/0003-metadata-api.md
```diff
@@ -0,0 +1,130 @@
...
+    type: enum(VIDEO|IMAGE|AUDIO),      // type of media
+    file: string,                       // media file url
+    loop: boolean,                      // if media wiil play in loop or not (only for audio and video)
+    autoplay: boolean                   // if media will play automatically or not
+  }
+}
+
+TransactionMeta {
+  id: string,                           // hash id of token or transaction
```
pedroferreira1, line 67: Suggestion: 'hash id of token or transaction ' -> 'transaction hash'

pedroferreira1 (approved):

PR #52, opened 2021-07-21
fix: clear obsolete node cache data

giovanecosta:
# TODO
- [ ] Clear current Redis data

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -82,6 +82,7 @@ jobs:
           REDIS_DB: 0
           TOKEN_METADATA_BUCKET: token-metadata-testnet
           CORS_ALLOWED_ORIGIN: https://explorer.testnet.hathor.network
+          NODE_CACHE_TTL: 5
       - name: Deploy Lambdas Mainnet
         if: ${{ needs.init.outputs.environment == 'mainnet' }}
         run: |
@@ -96,6 +97,7 @@ jobs:
           REDIS_DB: 0
           TOKEN_METADATA_BUCKET: token-metadata-mainnet
           CORS_ALLOWED_ORIGIN: https://explorer.hathor.network
+          NODE_CACHE_TTL: 5
       - name: Deploy Daemons Testnet
         if: ${{ needs.init.outputs.environment == 'testnet' }}
         run: |
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -82,7 +82,7 @@ jobs:
           REDIS_DB: 0
           TOKEN_METADATA_BUCKET: token-metadata-testnet
           CORS_ALLOWED_ORIGIN: https://explorer.testnet.hathor.network
-          NODE_CACHE_TTL: 5
+          NODE_CACHE_TTL: 30
       - name: Deploy Lambdas Mainnet
         if: ${{ needs.init.outputs.environment == 'mainnet' }}
         run: |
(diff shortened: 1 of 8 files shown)


pedroferreira1 (approved):
The PR for me is good, I was just curious about the ttl as 5 seconds. I feel is too fast to delete some data when the node might just lose some connection for a few seconds. Maybe increase to 30 or 60 seconds. What do you think? @giovanecosta @msbrogli 

Besides that I was thinking about the explorer itself. We keep polling this data when we are seeing a single peer, right? What happens if the peer data is deleted? The page will just become empty?

giovanecosta:
@pedroferreira1 30 seconds sounds fair enough for me.

Very good point on explorer. I've created an issue for that here: https://github.com/HathorNetwork/hathor-explorer/issues/93

.github/workflows/deploy.yml
```diff
@@ -82,6 +82,7 @@ jobs:
           REDIS_DB: 0
           TOKEN_METADATA_BUCKET: token-metadata-testnet
           CORS_ALLOWED_ORIGIN: https://explorer.testnet.hathor.network
+          NODE_CACHE_TTL: 5
```
msbrogli, line 85: 5 seconds? I feel we should wait at least 30 seconds before removing the data.

.github/workflows/deploy.yml
```diff
@@ -96,6 +97,7 @@ jobs:
           REDIS_DB: 0
           TOKEN_METADATA_BUCKET: token-metadata-mainnet
           CORS_ALLOWED_ORIGIN: https://explorer.hathor.network
+          NODE_CACHE_TTL: 5
```
msbrogli, line 100: Same thing here. 30 seconds maybe?

common/configuration.py
```diff
@@ -37,6 +37,7 @@ def default(cls) -> 'Environment':
 HATHOR_CORE_DOMAIN = config('HATHOR_CORE_DOMAIN', default=None)
 
 HATHOR_NODES = config('HATHOR_NODES', default='', cast=Csv())
+NODE_CACHE_TTL = config('NODE_CACHE_TTL', default=5)
```
msbrogli, line 40: Same here.

gateways/node_gateway.py
```diff
@@ -60,7 +60,7 @@ def save_node(self, id: str, node: Node) -> Union[bool, None]:
         :return: If saved successfuly or not
         :rtype: bool
         """
-        return self.cache_client.set(NODE_COLLECTION_NAME, id, node.to_dict())
+        return self.cache_client.set(NODE_COLLECTION_NAME, id, node.to_dict(), NODE_CACHE_TTL)
```
msbrogli, line 63: This won't fix the entries that are already there, right? Will we do a manual clean up?

msbrogli (changes requested):

gateways/node_gateway.py
giovanecosta: Yes, is in PR's TODO list

.github/workflows/deploy.yml
giovanecosta: Ok, we all agree with 30 then. Changed.

.github/workflows/deploy.yml
giovanecosta: changed

common/configuration.py
giovanecosta: changed

msbrogli (approved):

PR #54, opened 2021-07-29
fix: remove cache from node endpoints

giovanecosta:


pedroferreira1 (approved):

serverless.yml
```diff
@@ -78,10 +78,6 @@ functions:
             parameters: 
               paths: 
                 hash: true
-          caching:
```
msbrogli, line 81: Why are you removing cache?

msbrogli (approved):

serverless.yml
luislhl: I'm also curious about this.

giovanecosta: this is the objective of the PR

luislhl: But why it had to be removed?

luislhl: Ok, I have aligned this via Slack

PR #57, opened 2021-08-04
feat: report api

giovanecosta:
[Rendered Design](https://github.com/HathorNetwork/hathor-explorer-service/blob/71a00235e01ad46f62fc9cad5da49fd268e3257b/text/0005-send-report.md)

### DIFF
--- a/gateways/clients/report_client.py
+++ b/gateways/clients/report_client.py
@@ -1,5 +1,12 @@
 class ReportClient:
 
     def send(self, message: str) -> bool:
-        print(f"Report sent! Message: {message}")
+        """Send report
+
+        :param message: message to be sent
+        :type message: str
+        :return: if report was successful sent or not
+        :rtype: bool
+        """
+        print(f"Report sent! Message: {message}")  # TODO: THis line will be changed to real messenger provider
         return True
--- a/gateways/clients/report_client.py
+++ b/gateways/clients/report_client.py
@@ -1,0 +1,5 @@
+class ReportClient:
+
+    def send(self, message: str) -> bool:
+        print(f"Report sent! Message: {message}")
+        return True
--- a/text/0005-send-report.md
+++ b/text/0005-send-report.md
@@ -1,0 +1,67 @@
+- Feature Name: send_report_api
+- Start Date: 2021-08-02
+- RFC PR: (leave this empty)
+- Hathor Issue: (leave this empty)
+- Author: Giovane Costa <gigio.coder@gmail.com>
+
+# Summary
+[summary]: #summary
+
+Report API will receive reports from users about tokens, transactions, etc.
+
+# Motivation
+[motivation]: #motivation
+
+We are concerned about scammers, abuses, etc. As we have a metadata API that will carry information about tokens and
+transactions, we need a way where users can report tokens used in scams and respective transactions.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Report will be received as a POST and sent to a special email from hathor.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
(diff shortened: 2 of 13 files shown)


text/0005-send-report.md
```diff
@@ -0,0 +1,67 @@
...
+[reference-level-explanation]: #reference-level-explanation
+
+**Endpoint:**
+
+POST `/report` return `200` | `500`
+
+```
+body:
+{
+    type: str, // 'token' or 'transaction'
```
msbrogli, line 33: Should we have types of report as well? An initial list of choices might be: "impersonate" and "scam"?

text/0005-send-report.md
```diff
@@ -0,0 +1,67 @@
...
+    id: str,
+    decription: str
```
msbrogli, line 35: Should we put a max length here?

text/0005-send-report.md
```diff
@@ -0,0 +1,67 @@
...
+
+**Endpoint:**
+
+POST `/report` return `200` | `500`
+
+```
+body:
+{
+    type: str, // 'token' or 'transaction'
+    id: str,
```
msbrogli, line 34: How is the id calculated? I feel we shouldn't receive it here.

text/0005-send-report.md
msbrogli, line 33: Should we require a captcha to prevent spam? If yes, maybe it must have two steps.

text/0005-send-report.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None
+
+# Rationale and alternatives
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+We can use other way to receive reports instead of an email
```
msbrogli, line 56: I feel we should add a quick note about why we haven't chose alternative methods.

text/0005-send-report.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Report will be received as a POST and sent to a special email from hathor.
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+**Endpoint:**
```
msbrogli, line 26: It seems this is the Guide Level explanation. For the low level, I think we should explain what the API will do. 

I feel we should internally save an S3 file for each report, add a report counter to the token/transaction, and send an email. Maybe the email will be disabled in the future in case we start to receive too many reports.

msbrogli (changes requested):
I just reviewed the design.

text/0005-send-report.md
```diff
@@ -0,0 +1,67 @@
...
+# Unresolved questions
+[unresolved-questions]: #unresolved-questions
+
+None
+
+# Future possibilities
+[future-possibilities]: #future-possibilities
+
+We can store it in a database and have some automation as the volume can increase.
+It can be hard to us to manage all reports as it grows.
```
luislhl, line 67: Maybe we could use a third-party tool for this in the future. It would decrease the complexity for us.

Sentry had a User Feedback feature, for example: https://docs.sentry.io/api/projects/submit-user-feedback/

gateways/clients/report_client.py
```diff
@@ -0,0 +1,12 @@
...
+
+    def send(self, message: str) -> bool:
+        """Send report
+
+        :param message: message to be sent
+        :type message: str
+        :return: if report was successful sent or not
+        :rtype: bool
+        """
+        print(f"Report sent! Message: {message}")  # TODO: THis line will be changed to real messenger provider
```
luislhl, line 11: This will be done only in another PR?

PR #58, opened 2021-08-06
feat: token api

giovanecosta:
[Rendered Design](https://github.com/HathorNetwork/hathor-explorer-service/blob/5068afcf7b3d3c39d1499a9f6efc6a645e5b8440/text/0006-token-api.md)


text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
+- Feature Name: token_api
+- Start Date: 2021-08-04
+- RFC PR: (leave this empty)
+- Hathor Issue: (leave this empty)
+- Author: Giovane Costa <gigio.coder@gmail.com>
+
+# Summary
+[summary]: #summary
+
+Token API will provide token information by requesting `hathor-wallet-service`.
```
msbrogli, line 10: I guess we don't have to explain the low level in the summary. Maybe explain where it will be used.

text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+We need to retrieve token data without requesting the full node. Wallet service holds those data, so, we'll use it for now.
```
msbrogli, line 15: The motivation is to replace the full node's API so the explorer can scale as much as we want without hitting the full node direclty.

text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
...
+             timestamp: int
+        }
+    ]
+}
+```
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+None.
```
msbrogli, line 82: It depends on the wallet service.

text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
...
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+- We'll basically act as a proxy making requests to wallet service
+
+# Reference-level explanation
+[reference-level-explanation]: #reference-level-explanation
+
+**Endpoints:**
```
msbrogli, line 25: I guess these endpoints belong to the guide level explanation.

msbrogli, line 25: Maybe use `## Endpoints`.

text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
...
+
+GET `/token/:id` return `200` | `404`
```
msbrogli, line 27: Use `### GET ...`.

text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
...
+
+A request is made for Wallet service passing the same id
```
msbrogli, line 29: This explanation should be in the low level design. What is the API that will be accessed? Maybe add a link its documentation here.

tests/unit/usecases/test_list_tokens.py
```diff
@@ -0,0 +1,41 @@
+from unittest.mock import MagicMock
+
+from pytest import fixture
+
+from domain.tx.token import TokenList
+from tests.fixtures.token_factory import TokenListFactory
+from usecases.list_tokens import ListTokens
+
+
+class TestListTokens:
```
msbrogli, line 10: I don't feel we need tests as we are just proxying the response for the wallet service. What do you think?

tests/unit/usecases/test_list_token_transactions.py
```diff
@@ -0,0 +1,38 @@
+from unittest.mock import MagicMock
+
+from pytest import fixture
+
+from tests.fixtures.token_factory import TokenTransactionListFactory
+from usecases.list_token_transactions import ListTokenTransactions
+
+
+class TestListTokenTransactions:
```
msbrogli, line 9: As we are just doing a reverse proxy, do we still need tests?

tests/unit/gateways/clients/test_hathor_wallet_service_client.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_get(self, mocked_get):
+        mocked_get.return_value.status_code = 200
+        mocked_get.return_value.json = MagicMock(return_value={'success': True})
+        mocked_get.return_value.raise_for_status = MagicMock()
+
+        client = HathorWalletServiceClient('mydomain.com')
+
+        result = client.get('/some/path', {'page': 2})
+
+        mocked_get.assert_called_once_with('https://mydomain.com/some/path', params={'page': 2})
```
msbrogli, line 21: I like checking whether we are requesting the info from the wallet service.

tests/unit/gateways/clients/responses/hathor_wallet_service/test_hathor_wallet_service_token_response.py
```diff
@@ -0,0 +1,102 @@
...
+    HathorWalletServiceTokenTransactionsReponse,
+)
+from tests.fixtures.hathor_wallet_service_fixtures import (
+    HATHOR_WALLET_SERVICE_GET_TOKEN,
+    HATHOR_WALLET_SERVICE_GET_TOKEN_LIST,
+    HATHOR_WALLET_SERVICE_GET_TOKEN_TRANSACTION_LIST,
+)
+
+
+class TestHathorWalletServiceTokenResponse:
```
msbrogli, line 16: Same thing here. I don't know whether this test is useful since we are returning exactly the same thing as the wallet service.

tests/unit/domain/tx/test_token.py
```diff
@@ -40,7 +40,6 @@ def test_token_to_dict(self):
         assert token_dict['name'] == token.name
         assert token_dict['symbol'] == token.symbol
         assert token_dict['total_supply'] == token.total_supply
-        assert token_dict['transactions_count'] == token.transactions_count
```
msbrogli, line 43: Why did you remove it?

tests/unit/domain/tx/test_token.py
```diff
@@ -55,6 +54,5 @@ def test_token_from_dict(self):
         assert new_token.name == token.name
         assert new_token.symbol == token.symbol
         assert new_token.total_supply == token.total_supply
-        assert new_token.transactions_count == token.transactions_count
```
msbrogli, line 58: Why did you remove it?

serverless.yml
```diff
@@ -121,6 +121,32 @@ functions:
...
+        - 'handlers/list_tokens_handler.py'
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: tokens/
+          method: get
+          cors: true
+  
+  list_token_transactions_handler:
```
msbrogli, line 138: What is the timeout for this lambda? Should we use k8s instead?

serverless.yml
```diff
@@ -121,6 +121,32 @@ functions:
             ttlInSeconds: 15
             cacheKeyParameters:
               - name: request.path.hash
+  
+  list_tokens_handler:
```
msbrogli, line 125: What is the timeout for this lambda? Should we use k8s instead?

text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
...
+[motivation]: #motivation
+
+We need to retrieve token data without requesting the full node. Wallet service holds those data, so, we'll use it for now. 
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+- We'll basically act as a proxy making requests to wallet service
+
+# Reference-level explanation
```
msbrogli, line 22: I'm wondering whether we will just relay the response from the wallet service or not. I mean, should we have a well defined return type and map the Wallet Service's response into it? Or should we just be a reverse proxy and relay it?

I feel this is important to understand how this API might be affected by future changes in the Wallet Service's Token API.

handlers/list_tokens_handler.py
```diff
@@ -0,0 +1,37 @@
...
+        'size': event.query.get('size', None)
+    }
+
+    response = list_token.list(params)
+
+    if response is None:
+        raise Exception('not_found')
+
+    return {
+        "statusCode": 200,
```
msbrogli, line 32: What if `response.statusCode != 200`? Should we return 200 anyway?

handlers/list_token_transactions.py
```diff
@@ -0,0 +1,35 @@
...
+        'size': event.query.get('size', None)
+    }
+
+    response = list_token_transactions.list(event.path['hash'], params)
+
+    if response is None:
+        raise Exception('not_found')
+
+    return {
+        "statusCode": 200,
```
msbrogli, line 30: What if `response.statusCode != 200`? Should we return 200 anyway?

gateways/token_gateway.py
```diff
@@ -44,14 +56,85 @@ def get_token_metadata_from_s3(self, file: str) -> Union[TokenMetadata, None]:
...
+        :return: token domain class created from response
+        :rtype: Union[Token, None]
+        """
+        response = self.hathor_wallet_service_client.get(TOKEN_ENDPOINT.format(id=id), {})
 
         if response is None:
             return None
 
-        token_response = HathorCoreTokenResponse.from_dict(response)
+        token_response = HathorWalletServiceTokenResponse.from_dict(response)
```
msbrogli, line 71: I'm not sure whether we are just relaying the response or mapping to a well defined response type. In case we are relaying, I don't see why we would parse it.

gateways/clients/responses/hathor_wallet_service/hathor_wallet_service_token_response.py
```diff
@@ -0,0 +1,161 @@
...
+            return TokenList(
+                total=self.total,
+                tokens=[TokenListItem(
+                    id=token.id,
+                    name=token.name,
+                    symbol=token.symbol,
+                    timestamp=token.timestamp
+                ) for token in self.tokens]
+            )
+        except Exception:
```
msbrogli, line 160: It feel dangerous to catch `Exception`. Why is that? It would capture ANY exception generated inside it (even if your code is broken).

gateways/clients/responses/hathor_wallet_service/hathor_wallet_service_token_response.py
```diff
@@ -0,0 +1,161 @@
...
+            raise Exception('malformed_token_list')
```
msbrogli, line 161: Maybe raise a `ValueError`?

gateways/clients/responses/hathor_wallet_service/hathor_wallet_service_token_response.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        try:
+            return TokenTransactionList(
+                total=self.total,
+                transactions=[TokenTransactionListItem(
+                    id=transaction.id,
+                    timestamp=transaction.timestamp
+                ) for transaction in self.transactions]
+            )
+        except Exception:
```
msbrogli, line 102: It feel dangerous to catch `Exception`. Why is that? It would capture ANY exception generated inside it (even if your code is broken).

gateways/clients/responses/hathor_wallet_service/hathor_wallet_service_token_response.py
```diff
@@ -0,0 +1,161 @@
...
+            raise Exception('malformed_token_transaction_list')
```
msbrogli, line 103: Maybe raise `ValueError`?

gateways/clients/responses/hathor_wallet_service/hathor_wallet_service_token_response.py
```diff
@@ -0,0 +1,161 @@
...
+    name: str
+    symbol: str
+    timestamp: int
+    total_supply: int
+    can_mint: bool
+    can_melt: bool
+
+    @classmethod
+    def from_dict(cls, dikt: dict) -> 'HathorWalletServiceTokenResponse':
+        return from_dict(data_class=cls, data=dikt)
```
msbrogli, line 38: Does it validate type?

text/0006-token-api.md
msbrogli, line 22: Still in the low level design, how will it be deployed? Lambdas, right? I'm concerned that it can get really expensive if the Wallet Service's Token API hangs for any reason and we keep waiting for them. What would be a reasonable timeout here? Should we deploy it using k8s?

msbrogli (changes requested):
I reviewed the design and part of the code. I wasn't sure how to review the whole code since I still had some questions regarding the design.

domain/tx/token.py
```diff
@@ -126,3 +122,84 @@ def from_dict(cls, dikt: dict) -> 'Token':
...
+    total: int
+    tokens: List[TokenListItem]
+
+    def to_dict(self) -> dict:
+        """ Convert a TokenList instance into dict
+
+        :return: Dict representations of TokenList
+        :rtype: dict
+        """
+        return asdict(self)
```
jansegre, line 169: I feel like all of these `to_dict` and `from_dict` aren't really helping in any way. I'd be fine using `asdict` and `from_dict` directly, or at most making thin wrapper functions on a util module in case we want to customize some behavior. What do you think?

gateways/clients/responses/hathor_wallet_service/hathor_wallet_service_token_response.py
```diff
@@ -0,0 +1,161 @@
...
+
+from domain.tx.token import Token, TokenList, TokenListItem, TokenTransactionList, TokenTransactionListItem
+
+
+@dataclass
+class HathorWalletServiceTokenResponse:
+    """Response from hathor wallet service when requesting a token
+
+    :param id: id of token
+    :type id: str
```
jansegre, line 14: Are we generating a docs with these comments? If we are, is it possible to make it use the type annotations instead of having to duplicate the information with `:type`? If we aren't I feel like we have no feedback if we're making a mistake or not, and I wouldn't bother with `:type` at all, but the descriptions are nice.

text/0006-token-api.md
```diff
@@ -0,0 +1,92 @@
+- Feature Name: token_api
+- Start Date: 2021-08-04
+- RFC PR: (leave this empty)
+- Hathor Issue: (leave this empty)
+- Author: Giovane Costa <gigio.coder@gmail.com>
```
jansegre, line 5: Since this isn't in the RFC repo, I think we should remove this, because these fields are mostly specific to the RFC repo. And maybe also drop the `0006-` from the file name (I'm not sure if we have an RFC 0006 or not but it would certainly conflict with that).

jansegre (approved):

PR #70, opened 2021-09-02
tests: pdf type test failed because of library choosing path depending on category

pedroferreira1:


tests/fixtures/token_factory.py
```diff
@@ -21,8 +21,12 @@ class TokenNFTFactory(Factory):
     class Meta:
         model = TokenNFT
 
-    type = lazy_attribute(lambda o: fake.random_element(list(TokenNFTType)))
-    file = lazy_attribute(lambda o: f"http://{fake.domain_name()}{fake.file_path(category=o.type.value.lower())}")
```
r4mmer, line 25:
```python
file = lazy_attribute(lambda o: f"http://{fake.domain_name()}{fake.file_path(category=o.type.value.lower(), extension='pdf' if o.type.value.lower() == 'pdf' else None)}")
```

The lib used to generate will ignore `category` if `extension` is passed, this would fix the problem and not alter the factory logic.

r4mmer (changes requested):

tests/fixtures/token_factory.py
pedroferreira1: Can you see again please?

r4mmer (approved):

andreabadesso (approved):

PR #75, opened 2021-09-10
feat: Network data aggregator

r4mmer:
# Acceptance Criteria

- network data is always updated when node and peer data changes
- network endpoint

### DIFF
--- a/gateways/node_gateway.py
+++ b/gateways/node_gateway.py
@@ -1,9 +1,10 @@
-from typing import Callable, List, Union
+from typing import Callable, Dict, List, Union
 
 from common.configuration import DATA_AGGREGATOR_LAMBDA_NAME, NODE_CACHE_TTL
 from common.errors import ConfigError
+from domain.network.network import AggregatedNode, AggregatedPeer, Network
 from domain.network.node import Node
-from gateways.clients.cache_client import NODE_COLLECTION_NAME, CacheClient
+from gateways.clients.cache_client import NETWORK_COLLECTION_NAME, NODE_COLLECTION_NAME, CacheClient
 from gateways.clients.hathor_core_client import STATUS_ENDPOINT, HathorCoreAsyncClient
 from gateways.clients.lambda_client import LambdaClient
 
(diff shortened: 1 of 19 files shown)


gateways/node_gateway.py
```diff
@@ -77,10 +78,51 @@ def get_node(self, id: str) -> Union[Node, None]:
 
         return None
 
+    def save_network(self, network: Network) -> Union[bool, None]:
```
luislhl, line 81: Shouldn't we have created a `network_gateway.py` for those methods?

I think the pattern was to have one gateway for each class of the domain, right?

gateways/node_gateway.py
```diff
@@ -1,8 +1,9 @@
-from typing import Callable, List, Union
+from typing import Callable, Dict, List, Union
 
 from common.configuration import DATA_AGGREGATOR_LAMBDA_NAME, NODE_CACHE_TTL
```
luislhl, line 3: I know that this was not made in this PR, but I saw we are configuring the name of the Lambda using this Env Var `DATA_AGGREGATOR_LAMBDA_NAME`.

I think it should be better to just configure it directly in serverless.yml, I think we can make it inject the name of the lambda in the env var there.

usecases/collect_nodes_statuses.py
```diff
@@ -9,7 +9,7 @@
 class CollectNodesStatuses:
     """ Usecase class to collect nodes statuses
     """
-    def __init__(self, node_gateway: Union[NodeGateway, None] = None) -> None:
+    def __init__(self, node_gateway: Optional[NodeGateway] = None) -> None:
         self.node_gateway = node_gateway or NodeGateway()
 
     async def collect(self) -> None:
```
luislhl, line 15: As far as I understand, this method will trigger a call to the `node_data_aggregator` Lambda for each node, right?

Here Giovane told me that this happens in parallel: https://github.com/HathorNetwork/hathor-explorer-service/pull/3#pullrequestreview-673429229

And now we have added the aggregation logic in this same Lambda. So the aggregation logic will also be called for each node, and we will have parallel executions, if Giovane is right about parallelism.

Can't this lead to some nodes being missing from the final aggregated network? When comparing the old and new networks, you could be actually comparing one that was just saved by the other parallel executions of this Lambda?

Did you test this with several nodes to make sure this kind of problem doesn't happen?

gateways/node_gateway.py
r4mmer: Yes, but since the aggregation part will need to call or implement the `node_gateway` methods, one would import the other, so for now it seems best to keep this as a single class.
Or we could rename `node_gateway` to `network_gateway` since it interacts with the network.

gateways/node_gateway.py
r4mmer: created an issue for this #80

usecases/collect_nodes_statuses.py
r4mmer: My understanding was that this should not happen asynchronously, since all aggregations are not atomic, without lock and write to the same key on redis. I will test this further to check which steps are truly asynchronous.

luislhl:
This PR replaces the old one, right?
https://github.com/HathorNetwork/hathor-explorer-service/pull/42

usecases/collect_nodes_statuses.py
r4mmer: Correction, this will all happen in parallel and it is possible that some execution may skip a node update, but since the nodes are saved on their own redis keys the next execution will aggregate with the status on the node it missed on the last aggregation.
So, although some errors may occur, it will correct itself in a few seconds.

If we want to prevent it from happening at all we could invoke the save lambda in sync and when all node statuses are collected (and saved) we invoke the aggregate lambda.

luislhl: This seems good enough for me for now

luislhl (approved):

andreabadesso (approved):

PR #81, opened 2021-09-23
feat: node api for address balance and search

r4mmer:
# Acceptance criteria

- Implement the proxy api specified in HathorNetwork/internal-issues#42

# After deploy

- thin_wallet/address_search
- thin_wallet/address_balance

On the explorer full-node ALB redirect the calls to the apis above into the correspondent calls implemented here.


tests/fixtures/node_api_factory.py
```diff
@@ -0,0 +1,72 @@
...
+            transactions=None,
+            has_more=None,
+            total=None,
+        )
+
+    success = True
+    message = None
+    has_more = factory.Faker('boolean')
+    total = factory.Faker('random_int', min=1, max=999999)
+    transactions = factory.List([{}])  # TODO: mock txs
```
andreabadesso, line 72: Should we do this on this PR?

andreabadesso (changes requested):

tests/fixtures/node_api_factory.py
r4mmer: Added transaction model, factory and tests.

andreabadesso (approved):

domain/node_api/address_balance.py
```diff
@@ -0,0 +1,37 @@
...
+    received: int
+    spent: int
+
+
+@dataclass
+class AddressBalance:
+
+    success: bool
+    # The next fields are optional to parse when success is False
+    tokens_data: Optional[Dict[str, AddressBalanceTokenData]] = None
```
pedroferreira1, line 20: The key of the dict is the uid of the token?

serverless.yml
```diff
@@ -137,4 +137,40 @@ functions:
...
+  node_api_get_address_balance:
+    handler: handlers/node_api.get_address_balance
+    package:
+      patterns:
+        - 'handlers/node_api.py'
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: node_api/thin_wallet/address_balance
```
pedroferreira1, line 151: Can't we use only `node_api/address_balance`?

serverless.yml
```diff
@@ -137,4 +137,40 @@ functions:
...
+  node_api_get_address_search:
+    handler: handlers/node_api.get_address_search
+    package:
+      patterns:
+        - 'handlers/node_api.py'
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: node_api/thin_wallet/address_search
```
pedroferreira1, line 169: Can't we use only `node_api/address_search`?

handlers/node_api.py
```diff
@@ -0,0 +1,77 @@
...
+) -> dict:
+    node_api = node_api or NodeApi()
+    address = event.query.get("address")
+    if address is None:
+        raise Exception("invalid_parameters")
+
+    response = node_api.get_address_balance(address)
+    if response is None:
+        return {
+            "statusCode": 503,
```
pedroferreira1, line 24: Why 503?

usecases/node_api.py
```diff
@@ -0,0 +1,54 @@
...
+            result = self.node_api_gateway.get_address_search(
+                    address,
+                    count,
+                    page,
+                    hash,
+                    token)
+            if result is None:
+                return None
+            return result.to_dict()
+        except Exception as ex:
```
pedroferreira1, line 49: Can't we capture requests specific error for timeout?

usecases/node_api.py
```diff
@@ -0,0 +1,54 @@
...
+            if str(ex) == 'timeout':
```
pedroferreira1, line 50: Is there any better way to identify timeout?

gateways/node_api_gateway.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # /thin_wallet/address_search
+
+    def blacklist_address_search(self, address: str) -> Optional[bool]:
+        # No expiration on blacklist
+        return self.cache_client.set(
+                ADDRESS_SEARCH_BLACKLIST_COLLECTION_NAME,
+                address, 1)
+
+    def is_blacklisted_address_search(self, address: str) -> bool:
```
pedroferreira1, line 63: I'm thinking about this now. I know it's in the design but what are the benefits of separating two different blacklists? I believe we could have only one and both endpoints would return error. What do you think?

serverless.yml
r4mmer: We can, but when more apis are migrated to the explorer-service the `node-api` prefix will make it easier to identify which ones are a "proxy" to the full-node

handlers/node_api.py
r4mmer: The only error (besides timeout) that I found this api returning is 403 ("This path is forbidden in this server"). So a "Service Unavailable" seems more intuitive to return.
We could also do a refactor to make the `HathorCoreClient` always return the `requests` response, so we can bubble up the full-node response and return it to the client.

usecases/node_api.py
r4mmer: The `HathorCoreClient` raises an `Exception` with the text "timeout", instead of using the `requests` exception I prepared this `try..catch` to use a custom exception (i.e. `HathorCoreTimeout`).

I didn't create the `HathorCoreTimeout` exception yet because the custom errors PR is still in review.

gateways/node_api_gateway.py
r4mmer: Made them both use the same blacklist methods

serverless.yml
pedroferreira1: Yeah, I agree with `node_api`, I was just thinking about removing `thin_wallet`

domain/node_api/address_balance.py
pedroferreira1: I think we should add a comment about it

usecases/node_api.py
pedroferreira1: Ok. I guess we should wait until custom errors PR is merged to finish this PR and remove this global exception catch.

domain/node_api/address_balance.py
r4mmer: Done

serverless.yml
r4mmer: The thin_wallet is because i wanted to maintain the path structure from the full-node, to make it easier to proxy calls

usecases/node_api.py
r4mmer: It was merged and the PR has been updated.

gateways/node_api_gateway.py
```diff
@@ -0,0 +1,84 @@
...
+        :type address: str
+        """
+        value = self.hathor_core_client.get(
+                ADDRESS_BALANCE_ENDPOINT,
+                params={'address': address},
+                timeout=10,
+            )
+        if value is None:
+            return None
+        return AddressBalance.from_dict(value)
```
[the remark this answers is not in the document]
r4mmer: `{"success": false, "tokens_data": null, "total_transactions": null, "message": "Invalid 'address' parameter"}`

The null parameters were left because they are inconsequential, only `"success": False` and `message` are important.

usecases/node_api.py
r4mmer: Created a custom exception for this

serverless.yml
r4mmer: Changed to `node_api/address_balance`

Issue #83, opened 2021-10-05
[Design] Proxy full-node api


r4mmer:
# TODO

luislhl:
How the endpoints of the full-node were selected for this design? I think this should be described.

> Will be split in 2 endpoints

What do you mean by this? We will have 2 different endpoints in API Gateway both proxying to the same endpoint in the full-node?

> cache key: none (or block+tx)

I didn't understand if we will or will not use a cache key. Apparently it makes sense to use.

> /v1a/version

When this is called in the Explorer?

r4mmer:
> How the endpoints of the full-node were selected for this design? I think this should be described.

Of all endpoints the explorer uses, the only ones not included here are:
- `/v1a/push_tx`
    - should not be cached
- `/v1a/decode_tx`
    - there's an active issue to use the wallet-lib instead of this api
- `/v1a/thin_wallet/address_balance` and `/v1a/thin_wallet/address_search`
    - covered on another issue
- `/v1a/graphviz`
    - the use on the explorer wasn't standard, so I'm hesitant to include on a standardized proxy api

> What do you mean by this? We will have 2 different endpoints in API Gateway both proxying to the same endpoint in the full-node?

Yes, the `token` and `transaction` work differently depending on which parameters are used.
If you pass an id (for `tx_id` or `token_uid`) it will retrieve that transaction or token, but if id is not present it will assume a list operation.
For the transaction the list operation has some pagination parameters (and a filter by type).
For the token there are no parameters.

Since the parameters are so different the cache key cannot be the same for the get and list operations, which is why it makes more sense to split in 2 lambdas.

> I didn't understand if we will or will not use a cache key. Apparently it makes sense to use.

It does make sense to use, but on the explorer the block and tx parameters for this api are hard-coded, so all requests will retrieve the same amount of each.
If all requests retrieve the same amount, it makes sense to not use one as well, but i'm changing to using so the design won't be ambiguous.

> When this is called in the Explorer?

This api is called [here](https://github.com/HathorNetwork/hathor-explorer/blob/bdb38663bdf9003d1b85e9ab38847ae4349c618e/src/App.js#L62).
It's basically to set the network from the explorer to the same network of the backend (full-node) and check if the full-node version is allowed.

luislhl:
Approved

obiyankenobi:
> `/v1a/thin_wallet/token`
> - List tokens:
>   - cache 10 seconds
>   - cache key: none

[...]

PR #92, opened 2021-10-11
feat: node api proxy

r4mmer:
# Acceptance criteria

- Implements the api described on the design #83
- Does not alter parameters passed to the full-node and returns the exact response from the full-node

### DIFF
--- a/handlers/node_api.py
+++ b/handlers/node_api.py
@@ -16,6 +16,10 @@ def get_address_balance(
     _context: LambdaContext,
     node_api: Optional[NodeApi] = None
 ) -> dict:
+    """Get the token balance of a given address.
+
+        *IMPORTANT: Any changes on the parameters should be reflected on the `cacheKeyParameters` for this method.
+    """
     node_api = node_api or NodeApi()
     address = event.query.get("address")
     if address is None:
@@ -38,6 +42,10 @@ def get_address_search(
     _context: LambdaContext,
     node_api: Optional[NodeApi] = None
 ) -> dict:
+    """Get a paginated list of transactions for a given address.
+
+        *IMPORTANT: Any changes on the parameters should be reflected on the `cacheKeyParameters` for this method.
+    """
     node_api = node_api or NodeApi()
 
     address = event.query.get("address")
@@ -69,6 +77,10 @@ def get_version(
     _context: LambdaContext,
     node_api: Optional[NodeApi] = None
 ) -> dict:
+    """Get the node version settings.
+
+        *IMPORTANT: Any changes on the parameters should be reflected on the `cacheKeyParameters` for this method.
+    """
     node_api = node_api or NodeApi()
 
     response = node_api.get_version()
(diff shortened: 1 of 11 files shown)


handlers/node_api.py
```diff
@@ -60,18 +54,201 @@ def get_address_search(
...
+@ApiGateway()
+def get_dashboard_tx(
+    event: ApiGatewayEvent,
+    _context: LambdaContext,
+    node_api: Optional[NodeApi] = None
+) -> dict:
+    node_api = node_api or NodeApi()
+
+    block = event.query.get("block")
+    tx = event.query.get("tx")
```
luislhl, line 94: Maybe we should add warnings here that any new query parameter that someone eventually adds here will have to be added to serverless.yml's cacheKeys as well.

It seems very easy to make this kind of mistake.

handlers/node_api.py
```diff
@@ -60,18 +54,201 @@ def get_address_search(
...
+    hash = event.query.get("hash")
+    page = event.query.get("page")
+    timestamp = event.query.get("timestamp")
+
+    if id is None or count is None:
+        raise ApiError("invalid_parameters")
+
+    if hash is not None and (page is None or timestamp is None):
+        # If hash exists, it"s a paginated request and page is required
+        raise ApiError("invalid_parameters")
```
luislhl, line 152: Wouldn't it be better if we just proxy everything to the fullnode and let it perform the validation?

Otherwise we will need to update here every time someone updates the validation rules in hathor-core.

handlers/node_api.py
```diff
@@ -20,18 +22,10 @@ def get_address_balance(
...
-            "body": json.dumps({"error": "unknown_error"}),
-            "headers": {
-                "Content-Type": "application/json"
-            }
-        }
 
     return {
         "statusCode": 200,
-        "body": json.dumps(response),
+        "body": json.dumps(response or UNKNOWN_ERROR_MSG),
```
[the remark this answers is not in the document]
andreabadesso: I think that the fullnode returns status code 200 for all responses, including errors

andreabadesso (approved):
Other than @luislhl comments, this looks good to me

usecases/node_api.py
```diff
@@ -41,3 +41,39 @@ def get_address_search(
         except HathorCoreTimeout:
             self.node_api_gateway.blacklist_address(address)
             return ADDRESS_BLACKLIST_RESPONSE
```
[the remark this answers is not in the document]
r4mmer: Yes, the previous response was "This address is blacklisted" but the reason of the blacklist is a timeout due to too many transactions. This message was changed because it is showed to the user on the frontend so it should reflect the actual reason.

handlers/node_api.py
r4mmer: Done on the docstrings

handlers/node_api.py
r4mmer: Maybe, but filtering out invalid parameters may lessen the burden on the full-node. If this is needless we can remove easily on another PR.

usecases/node_api.py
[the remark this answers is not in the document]
luislhl: > but the reason of the blacklist is a timeout due to too many transactions

I'm not sure I understand this sentence.

I brought an example from line 18 in this file to make my point clearer:
```
        if self.node_api_gateway.is_blacklisted_address(address):
            return ADDRESS_BLACKLIST_RESPONSE
```

I don't understand why in this case we return `Timeout due to too many transaction` to the user.

handlers/node_api.py
luislhl: Ok, I see your point

luislhl (approved):

PR #106, opened 2021-12-03
fix: daemon error on status timeout

r4mmer:
# Summary

During some calls the node would timeout the response (with status 504), or return a status 429 and although aiohttp does not raise an exception for this it does raise when trying to read a json from a mime-type html response (which is the case on both errors).
The error treatment of the HathorCoreAsyncClient is to send the error to the callback, but the callback for the daemon does not have an error treatment.

# Acceptance criteria

- Ignore errors when collecting node status
- Log responses and errors from hathor core to make errors like this easier to spot

### DIFF
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -30,6 +30,7 @@ class HathorCoreAsyncClient:
         :type domain: str, optional
         """
         self.domain = domain or HATHOR_CORE_DOMAIN
+        self.log = logger.new(client="async")
 
     async def get(self, path: str, callback: Callable[[dict], None], params: Optional[dict] = None) -> None:
         """Make a get request async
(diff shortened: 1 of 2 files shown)


luislhl (approved):

gateways/clients/hathor_core_client.py
```diff
@@ -46,8 +47,14 @@ async def get(self, path: str, callback: Callable[[dict], None], params: Optiona
         try:
             async with aiohttp.ClientSession() as session:
                 async with session.get(url, params=params) as response:
+                    self.log.info(
```
pedroferreira1, line 50: How does this `await` works? I guess it shouldn't matter to have 2 awaits, right? Or should we have the await first, like this

```
awaited_response = await response

body=awaited_response.text())
callback(awaited_response.json())
```

To be honest I'm not sure how this await works, it's just a question

pedroferreira1 (approved):

gateways/clients/hathor_core_client.py
r4mmer: Well, the basic is that any async method returns a coroutine and to get the result of that coroutine you need to `await` it. The response being awaited does not mean that the `text` and `json` methods don't need to be awaited since both of them are async methods.
I think these methods are async because some http response bodies can be chunked, which would make the headers and status of the request available first (status and headers are not async methods) and the body available after waiting for the responder to push the data (which is being waited on asynchronously).

PR #114, opened 2022-01-04
chore: move testnet node to golf

r4mmer:


.github/workflows/deploy.yml
```diff
@@ -74,8 +74,8 @@ jobs:
           npx make deploy-lambdas-testnet
         env:
           API_PORT: 3001
-          HATHOR_CORE_DOMAIN: node1.foxtrot.testnet.hathor.network
-          HATHOR_NODES: node1.foxtrot.testnet.hathor.network
+          HATHOR_CORE_DOMAIN: node1.golf.testnet.hathor.network
+          HATHOR_NODES: node1.golf.testnet.hathor.network
```
luislhl, line 78: Maybe we should change to `node1.testnet.hathor.network`, so we don't need to update here every time we replace the testnet by a new one.

r4mmer: Makes sense.

luislhl (approved):

pedroferreira1 (approved):

PR #116, opened 2022-01-11
fix: filter unused known_peers

r4mmer:
# Acceptance criteria

- Filter `known_peers` before sending to lambda so it doesn't reach the data limit of the lambda.
- Have only the peers that are also in `connected_peers` on the `known_peers` (others are ignored on the explorer)

### DIFF
--- a/usecases/collect_nodes_statuses.py
+++ b/usecases/collect_nodes_statuses.py
@@ -1,4 +1,5 @@
 from typing import Optional
+from copy import copy
 
 from common.configuration import HATHOR_NODES
 from common.logging import get_logger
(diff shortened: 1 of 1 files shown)


usecases/collect_nodes_statuses.py
```diff
@@ -29,4 +30,11 @@ def _send(self, data: dict) -> None:
             self.log.warning("collect_status_error", error=data['error'])
             return
         node = Node.from_status_dict(data)
+        # XXX: a big enough known_peers list can reach the limit of data passed to a lambda.
```
pedroferreira1, line 33: I think we should reference here the issues that are handling this problem

usecases/collect_nodes_statuses.py
```diff
@@ -29,4 +30,11 @@ def _send(self, data: dict) -> None:
...
+        # This filters the unnecessary peer ids.
+        old_known_peers = copy(node.known_peers)
```
pedroferreira1, line 35: Do we need this copy? If it's not needed we should remove it so we don't use this amount of ram in the lambda

pedroferreira1 (approved):

luislhl (approved):

usecases/collect_nodes_statuses.py
r4mmer: Added

PR #121, opened 2022-02-04
feat: enhance CORS validation to support multiple origins (Closes #119)

### LINKED ISSUE #119: Add support for multiple URLs for CORS

lucas3003:
### Description

Customers of testnet were experiencing CORS problems when trying to access from `https://explorer.golf.hathor.network`. The temporary solution was to redirect all requests to `https://explorer.testnet.hathor.network`. With this change, `golf` will start to work again (with no redirects needed) and we will be able to change the rules using regex.

### Acceptance Criteria

Customers accessing  `https://explorer.golf.hathor.network` must be able to access explorer without being redirected. Also, current customers accessing `https://explorer.testnet.hathor.network` and `https://explorer.hathor.network` must **NOT** be impacted.

This change was tested both locally and uploading the stack to an AWS account.


.github/workflows/deploy.yml
```diff
@@ -96,7 +96,7 @@ jobs:
           REDIS_PORT: 6379
           REDIS_DB: 0
           METADATA_BUCKET: hathor-explorer-metadata-mainnet
-          CORS_ALLOWED_ORIGIN: https://explorer.hathor.network
+          CORS_ALLOWED_REGEX: https?://([a-z0-9])*[.]*hathor[.]network
```
luislhl, line 99: This regex is different from the above. 
`https?://([a-z0-9])*[.]*hathor[.]network`
`https?://([a-z0-9]+[.])*hathor[.]network`

luislhl: I think I get it now. You made them different because in mainnet the address is only `explorer.hathor.netowrk`, right?

But we actually have this one too that I think we would want to support: `https://explorer.mainnet.hathor.network/`

pedroferreira1: I guess it will be supported. I think even `https://hathor.network` will be supported

utils/wrappers/aws/api_gateway.py
```diff
@@ -41,11 +42,13 @@ def wrapper(event: dict, context: LambdaContext, *args: Any, **kwargs: Any) -> d
...
 
-            headers = {}
+            requestHeaders = event.get('headers', {})
+            origin = requestHeaders.get('origin', requestHeaders.get('Origin', None))
 
-            if CORS_ALLOWED_ORIGIN:
-                headers = {
-                    'Access-Control-Allow-Origin': CORS_ALLOWED_ORIGIN,
+            responseHeaders = {}
+            if origin is not None and re.match(CORS_ALLOWED_REGEX, origin):
```
r4mmer, line 49: Do we use any requests that do not include the `Origin` header? And not having cors headers for not allowed origins work as intended?

Also, should we not run the handler if the origin does not match?

.github/workflows/deploy.yml
r4mmer: Maybe `https?://([a-z0-9]*\.){0,5}hathor\.network` which allows the main domain and any subdomain up to 5 parts (I think we only use 3).

r4mmer (changes requested):

utils/wrappers/aws/api_gateway.py
lucas3003: Do we use any requests that do not include the Origin header? And not having cors headers for not allowed origins work as intended?

- I don't think we have those requests. Only node_data_aggregator handler does not have the ApiGateway() decorator.

Also, should we not run the handler if the origin does not match?

- You mean ,we should not do anything if CORS does not match?

common/configuration.py
```diff
@@ -66,6 +66,6 @@ def default(cls) -> 'LogRenderer':
 
 METADATA_BUCKET = config('METADATA_BUCKET', default=None)
 
-CORS_ALLOWED_ORIGIN = config('CORS_ALLOWED_ORIGIN', default=None)
+CORS_ALLOWED_REGEX = config('CORS_ALLOWED_REGEX', default=None)
```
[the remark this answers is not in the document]
lucas3003: I will put it here

.github/workflows/deploy.yml
lucas3003: I will use Andre suggestions as it contemplates

https://explorer.hathor.network
https://explorer.mainnet.hathor.network
https://hathor.network

.github/workflows/deploy.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           REDIS_PORT: 6379
           REDIS_DB: 0
           METADATA_BUCKET: hathor-explorer-metadata-testnet
-          CORS_ALLOWED_ORIGIN: https://explorer.testnet.hathor.network
+          CORS_ALLOWED_REGEX: https?://([a-z0-9]+[.])*hathor[.]network
```
r4mmer, line 84: Also change the regex here.

r4mmer (changes requested):

pedroferreira1 (approved):

r4mmer (approved):

PR #122, opened 2022-02-08
Make it easier to run hathor-explore-service locally

lucas3003:
### Description

Today, running hathor-explorer-service is a trial-and-error process. The developer must know how serverless work, and how to install all dependencies.
 
### Acceptance Criteria
Now, just running `docker-compose up -d` will be enough to make explorer-service run on port 3001.


luislhl:
Renaming the Dockerfile will break the deploy job in Github Actions.

We need to use the new name where need in the `Makefile`

Dockerfile_Service
```diff
@@ -0,0 +1,25 @@
+FROM nikolaik/python-nodejs:python3.8-nodejs16
+
+RUN apt-get update
+
+# Install serverless
+RUN npm install -g serverless && \
+    npm install -g serverless-offline
```
luislhl, line 7: These dependencies are defined in package.json. Why don't we install from it?

Same thing for the `RUN serverless plugin install -n serverless-s3-local` down there.

luislhl: Another thing is that you didn't lock the version of the packages to be installed, which is something we would get by using package.json

serverless-local.yml
```diff
@@ -0,0 +1,391 @@
...
+        - 'handlers/get_network_handler.py'
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: network/
+          method: get
+          cors: true
+
+  node_data_aggregator_handler:
```
luislhl, line 102: This Lambda is called by the Daemon based on its name in https://github.com/HathorNetwork/hathor-explorer-service/blob/7c90273201fc5390874c61b0eeb215f6d891efa4/gateways/node_gateway.py#L51

I saw that you used `.env.example` as env for the containers, and there we have:

```
LAMBDA_INVOKE_URL=http://localhost:3002/
DATA_AGGREGATOR_LAMBDA_NAME=hathor-explorer-service-dev-node_data_aggregator_handler
```

Did you test if this Lambda is correctly called by the daemon when running them in docker-compose?

I would assume it's not, because of the `localhost` address.

serverless-local.yml
```diff
@@ -0,0 +1,391 @@
...
+# The `service` block is the name of the service
+service: hathor-explorer-service
+useDotenv: true
+# The `provider` block defines where your service will be deployed
+provider:
+  name: aws
+  runtime: python3.8
+  region: ${env:AWS_DEFAULT_REGION}
+  lambdaHashingVersion: 20201221
+  environment:
```
luislhl, line 16: Maybe duplicating the serverless.yml file is not a good thing. They could become different if someone is not careful when changing them.

What problem did you have with the `provider.vpc` config that you had to remove?

Dockerfile_Service
```diff
@@ -0,0 +1,25 @@
...
+WORKDIR /app
+COPY . /app
+
+# Install dependencies
+RUN poetry config virtualenvs.create false \
+  && poetry install --no-dev --no-interaction --no-ansi 
+
+RUN npm install
+
+ADD serverless-local.yml serverless.yml
```
luislhl, line 18: Why do you need to add both?

Dockerfile_Service
```diff
@@ -0,0 +1,25 @@
...
+
+# Install necessary plugins
+RUN serverless plugin install -n serverless-s3-local
+
+# Run the service
+EXPOSE 3001
+CMD [ "npm", "start" ]
```
luislhl, line 25: Where is this `start` command coming from, since you added `package.json` to `.dockerignore`? From `package-lock.json`?

And which serverless file will it use, `serverless-local.yml` or `serverless.yml`?

luislhl (dismissed):
See my previous comments

Dockerfile_Service
```diff
@@ -0,0 +1,25 @@
...
+# Install dependencies
+RUN poetry config virtualenvs.create false \
+  && poetry install --no-dev --no-interaction --no-ansi 
+
+RUN npm install
+
+ADD serverless-local.yml serverless.yml
+
+# Install necessary plugins
+RUN serverless plugin install -n serverless-s3-local
```
r4mmer, line 21: This should be already on the `package.json`

Dockerfile_Service
```diff
@@ -0,0 +1,25 @@
+FROM nikolaik/python-nodejs:python3.8-nodejs16
+
+RUN apt-get update
+
+# Install serverless
+RUN npm install -g serverless && \
+    npm install -g serverless-offline
+
+WORKDIR /app
+COPY . /app
```
r4mmer, line 10: Not really an issue, but maybe we could install dependencies first then add the code. This makes it so you don't have to install dependencies again when building locally after a change on the source code.

A good order would be:
- add `package.json` and lock file
- run `npm install`
- add `pyproject.toml` and lock file
- run poetry install
- add source code

Dockerfile_Service
```diff
@@ -0,0 +1,25 @@
+FROM nikolaik/python-nodejs:python3.8-nodejs16
+
+RUN apt-get update
+
+# Install serverless
+RUN npm install -g serverless && \
```
r4mmer, line 6: This should be on the `package.json`

r4mmer (changes requested):
Maybe change the name for `Dockerfile_Service` and `Dockerfile_Daemons` to `daemon.Dockerfile` and `service.Dockerfile`. Some editors recognize the `.Dockerfile` extension as a Dockerfile and enable syntax highlight for them.

And what is the purpose of the extra serverless file? Is there no way to reconcile our serverless file to run locally?

Dockerfile_Service
lucas3003: Acually package.json. is being allowed instead of removed. We will use serverless-local.yml

Dockerfile_Service
luislhl: Sorry, you're right

docker-compose.yml
```diff
@@ -1,12 +1,23 @@
...
   daemons:
-    build: .
+    build:
+      context: .
+      dockerfile: Dockerfile_Daemons 
     stdin_open: true
-    network_mode: host
     env_file:
-      - .env
+      - .env.example
```
r4mmer, line 9: We should keep `.env`, since we may want to change the variables for local development. And let the `.env.example` only as an example.
Maybe add "run `cp .env.example .env` before running locally" on `README.md`

r4mmer (approved):

PR #130, opened 2022-02-18
Fix CORS problem due to API GW caching

lucas3003:
### Description

If the user access ```https://explorer.hathor.network```, it will work normally. However, if after that the user tries to access ```https://explorer.mainnet.hathor.network``` he/she will see a CORS problem. This happens because the first request is cached and returned on the second request. The browser identifies that the ```Origin``` in the request is different than the ```Access-Control-Allow-Origin``` in the response and shows the CORS error message to the user.

To solve that, the ```Origin``` was introduced as a cache key on ```node_api_get_version``` endpoint, which is the API that is returning CORS error.

### Testing

Tests were performed using Postman and deploying a local stack of Explorer Service on AWS. Now, when the origin changes (And matches the Regex), the ```Access-Control-Allow-Origin``` returns the correct value.

### Acceptance criteria

User must not see CORS error message on any domain that matches the regex.

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -260,6 +260,8 @@ functions:
           caching:
             enabled: true
             ttlInSeconds: 3600
+            cacheKeyParameters:
+              - name: request.header.Origin
 
   node_api_get_dashboard_tx:
     handler: handlers/node_api.get_dashboard_tx
--- a/serverless.yml
+++ b/serverless.yml
@@ -125,6 +125,9 @@ functions:
             parameters:
               paths:
                 hash: true
+          caching:
+            cacheKeyParameters:
+              - name: request.header.origin
 
   list_nodes_handler:
     handler: handlers/list_nodes_handler.handle
@@ -138,6 +141,9 @@ functions:
           path: node/
           method: get
           cors: true
+          caching:
+            cacheKeyParameters:
+              - name: request.header.origin
 
   get_network_handler:
     handler: handlers/get_network_handler.handle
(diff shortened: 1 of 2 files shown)


r4mmer (approved):

serverless.yml
```diff
@@ -257,6 +257,8 @@ functions:
           caching:
             enabled: true
             ttlInSeconds: 3600
+            cacheKeyParameters:
+              - name: request.header.Origin
```
luislhl, line 261: Are you sure the problem will only occur in this endpoint?

There are other endpoints called by the Explorer, they will still be caching the CORS headers and may present the same problem, right?

r4mmer: @luislhl The problem will occur on all endpoints since we have cors enabled on all of them.

We should check that we can add the `cacheKeyParameters` on the global `apiGatewayCaching` and that it will not conflict with the other apis that have their own `cacheKeyParameters`. If there is a conflict, we can just add the header on each api.

serverless.yml
```diff
@@ -125,6 +125,9 @@ functions:
             parameters:
               paths:
                 hash: true
+          caching:
```
luislhl, line 128: This won't enable caching on this endpoint unless we add `enabled: true`, right?

I'm not sure if you wanted to enabled caching, but I think we shouldn't enable in endpoints that didn't have it enabled before, unless you have a good reason for this.

If you didn't want to enable caching, but only make it future-proof (in case someone enables caching in the future), then I think we should just leave a comment to make it explicit that the current configuration doesn't enable caching, because someone reading the code will assume caching is enabled by this (at least until further looking into the plugin docs)

Does it make sense?

lucas3003: You're right. We might not put caching properties on APIs that do not have cache enabled. I am sending a new version correcting this part.

luislhl (approved):

PR #139, opened 2022-04-02
chore: stop diff early

r4mmer:
# Summary

On the data aggregator lambda, the [deepdiff](https://github.com/seperman/deepdiff) lib is used to check if there is any difference between the network data that was aggregated and the one saved on cache. If there is a difference we save the most recent one.

The network data is aggregated from node statuses of the nodes, and some reach over 32Kb of data (after filtering unnecessary data, before filtering it would be over 1Mb) this makes the deepdiff take more time to find all differences on every request.

The proposed solution is to stop the deepdiff when it finds the first difference.

# Acceptance criteria

- [x] Stop the deepdiff of the network data aggregator when it finds the first difference.
- [x] Increase the data aggregator lambda timeout
- [x] Log where needed for better understanding when a lambda timeout occurs

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -154,6 +154,7 @@ functions:
 
   node_data_aggregator_handler:
     handler: handlers/node_data_aggregator.handle
+    timeout: 30
     package:
       patterns:
         - 'handlers/node_data_aggregator.py'
--- a/usecases/aggregate_node_data.py
+++ b/usecases/aggregate_node_data.py
@@ -1,12 +1,25 @@
 from typing import Union
 
 from deepdiff import DeepDiff
+from deepdiff.model import DiffLevel
 
 from gateways.node_gateway import NodeGateway
 
 
-class AggregateNodeData:
+class EarlyStopDiff:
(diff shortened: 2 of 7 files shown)


serverless.yml
```diff
@@ -154,6 +154,7 @@ functions:
 
   node_data_aggregator_handler:
     handler: handlers/node_data_aggregator.handle
+    timeout: 30
```
luislhl, line 157: Why are you raising this to 30s ?

usecases/aggregate_node_data.py
```diff
@@ -1,24 +1,50 @@
...
-class AggregateNodeData:
 
+class EarlyStopDiff:
+    """ Stop the diff when we find an error
+
+    From the docs example at:
+    https://zepworks.com/deepdiff/5.7.0/custom.html#custom-operators
+    """
+    def match(self, level: DiffLevel) -> bool:
+        return True
```
luislhl, line 20: What does this `match` function mean for the lib?

lucas3003 (approved):

PR #145, opened 2022-04-18
feat: add throttling to tokens api

lucas3003:
### Acceptance Criteria

Throttle tokens API to prevent resource exhaustion

The throttle values were created based on the traffic that was observed on Transactions API.

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -42,6 +42,7 @@ plugins:
   - serverless-python-requirements
   - serverless-api-gateway-caching
   - serverless-plugin-aws-alerts
+  - serverless-api-gateway-throttling
 
 package:
   individually: true
@@ -212,6 +213,9 @@ functions:
               - name: request.querystring.order
               - name: request.querystring.search_after
               - name: request.header.origin
+          throttling:
+            maxRequestsPerSecond: 1000
+            maxConcurrentRequests: 1000
 
   get_dag_metadata_handler:
     handler: handlers/get_dag_metadata_handler.handle
--- a/serverless.yml
+++ b/serverless.yml
@@ -68,6 +68,11 @@ custom:
   apiGatewayCaching:
     enabled: true
     ttlInSeconds: 300
+  apiGatewayThrottling:
+    # Apply to all http endpoints, unless overridden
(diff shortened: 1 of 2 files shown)


serverless.yml
```diff
@@ -212,6 +213,9 @@ functions:
               - name: request.querystring.order
               - name: request.querystring.search_after
               - name: request.header.origin
+          throttling:
+            maxRequestsPerSecond: 1000
+            maxConcurrentRequests: 1000
```
luislhl, line 218: I think we should take the opportunity to add throttling in all the Lambdas, not only this one.

Also, shouldn't we configure lower numbers here? According to https://www.serverless.com/plugins/serverless-api-gateway-throttling, the account-level limits are 10,000 and 5,000 for these settings. 1,000 seems too high for a single Lambda compared to the account, and too high compared to our current usage levels.

Maybe 50/s and 25 concurrent per Lambda would be better. What do you think?

lucas3003: Yes, I think we can use this opportunity to set throttling for all lambdas, not just this one. I agree with 50/second and 25 concurrent per Lambda. I will update this PR.

r4mmer: @luislhl do you think we can have some type of dashboard (on cloudwatch or grafana) to check if we are close to reaching this throttling limit? This way we could check which functions are more/less used and adjust the limit accordingly in the future.

luislhl: I thought about it as well. @r4mmer 

We would have it for the Lambdas already. The problem is the number of Lambda executions may be lower than the number of requests in API Gateway, because of cache. But it could give us some notion.

For API Gateway, AWS does not provide metrics aggregated by endpoint, only global metrics.

So I don't really know yet how we could do this, I'll take a further look.

But I know there is a way to have an alert when some request returns 429, because it is logged in API Gateway logs (when they are enabled). We could get this log and build a metric out of it, then an alert.

Maybe we can use logs to have the number of requests in each endpoint as well, but this depends on what they log exactly. I'll have a better notion when our logs are enabled.

luislhl (approved):

r4mmer (approved):

PR #148, opened 2022-04-19
feat: add transaction timestamp to Tokens API response

lucas3003:
### Acceptance Criteria

Adds transaction_timestamp to the Tokens API response, so the user can sort by oldest to newest tokens, or vice-versa.

Considerations:
- When passing a keyword, the date type must be passed on sort structure, e.g. `name.keyword` or `symbol.keyword`. However, when a `long` type is passed, like `transaction_timestamp` field, we should not append `.long`. This is why the data type structure was changed on this PR.

### DIFF
--- a/utils/elastic_search/elastic_search_utils.py
+++ b/utils/elastic_search/elastic_search_utils.py
@@ -29,22 +29,10 @@ class ElasticSearchUtils:
             sort_by = 'id'
 
         sort_fields_data_type = {
-            'transaction_timestamp': {
-                'type': 'long',
-                'append_data_type_on_sort': False
-            },
-            'id': {
-                'type': 'keyword',
-                'append_data_type_on_sort': True
-            },
-            'name': {
-                'type': 'keyword',
-                'append_data_type_on_sort': True
-            },
-            'symbol': {
-                'type': 'keyword',
-                'append_data_type_on_sort': True
-            }
+            'transaction_timestamp': 'long',
+            'id': 'keyword',
+            'name': 'keyword',
+            'symbol': 'keyword'
         }
 
         if not order:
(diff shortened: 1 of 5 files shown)


utils/elastic_search/elastic_search_utils.py
```diff
@@ -23,27 +23,51 @@ def build_search_query(self, search_text: str, sort_by: str, order: str, search_
...
             order = 'asc'
 
         primary_sort_key = {}
-        primary_sort_key[sort_by+'.'+sort_fields_data_type[sort_by]] = order
+
+        sort_by_complement = ''
+        if sort_fields_data_type[sort_by]['append_data_type_on_sort']:
+            sort_by_complement = '.'+str(sort_fields_data_type[sort_by]['type'])
+
+        primary_sort_key[sort_by+sort_by_complement] = order
```
r4mmer, line 59: Do we have to append the data type on any type other than `keyword`?
If not this would feel a bit awkward and we could just append when the type is `keyword`.

r4mmer (changes requested):

utils/elastic_search/elastic_search_utils.py
lucas3003: I am not sure if other data types will need to be appended or not, this is why I wanted to make as flexible as possible. Do you think we can keep this way?

r4mmer (approved):

utils/elastic_search/elastic_search_utils.py
```diff
@@ -38,12 +39,17 @@ def build_search_query(self, search_text: str, sort_by: str, order: str, search_
...
+        sort_by_complement = '.keyword' if sort_fields_data_type[sort_by] == 'keyword' else ''
+        primary_sort_key[sort_by+sort_by_complement] = order
+
         sort_order.remove(sort_by)
 
         tie_break_sort_order = sort_order.pop(0)
         tie_break_sort_key = {}
-        tie_break_sort_key[tie_break_sort_order+'.'+sort_fields_data_type[tie_break_sort_order]] = 'asc'
+
+        sort_by_complement = '.keyword' if sort_fields_data_type[tie_break_sort_order] == 'keyword' else ''
```
pedroferreira1, line 51: This is already used twice. Maybe we could create a util method in this same file that gets the `sort_by_complement` from a `sort_by` parameter

pedroferreira1 (approved):

PR #157, opened 2022-05-25
feat: AddressBalance list and query API

andreabadesso:
# Acceptance criteria

- We should have an API to list token balances for a given `token_id`
- We should have an API to fetch token information (transaction count and address count) for a given `token_id`


gateways/token_balances_api_gateway.py
```diff
@@ -0,0 +1,103 @@
...
+        elastic_search_client: Optional[Elasticsearch] = None,
+    ) -> None:
+        self.elastic_search_client = ElasticSearchClient(elastic_index=ELASTIC_TOKEN_BALANCES_INDEX,
+                                                         client=elastic_search_client)
+
+    def get_token_information(self, token_id: str) -> dict:
+        """Retrieve total number of addresses and transactions for a given token_id
+        """
+
+        body = {
```
lucas3003, line 33: index must be informed

gateways/clients/elastic_search_client.py
```diff
@@ -7,19 +7,22 @@
...
+    def __init__(self, elastic_index: str, client: Optional[Elasticsearch]) -> None:
         """Client to make async requests to ElasticSearch, using Cloud ID and Elastic Password
         """
 
         if client:
             self.client = client
         else:
             self.client = Elasticsearch(
-                cloud_id=ELASTIC_CLOUD_ID,
+                ELASTIC_CLOUD_ID,
```
lucas3003, line 18: I think it wont work without the `cloud_id`

gateways/token_balances_api_gateway.py
andreabadesso: Fixed in 0b2b67d176977d71660743283b3f0e4f3fe7af1b

common/configuration.py
```diff
@@ -76,3 +76,4 @@ def default(cls) -> 'LogRenderer':
 ELASTIC_INDEX = config('ELASTIC_INDEX', default=None)
 ELASTIC_RESULTS_PER_PAGE = config('ELASTIC_RESULTS_PER_PAGE', default=10)
 ELASTIC_SEARCH_TIMEOUT = config('ELASTIC_SEARCH_TIMEOUT', default=25)
+ELASTIC_TOKEN_BALANCES_INDEX = config('ELASTIC_TOKEN_BALANCES_INDEX', default=None)
```
lucas3003, line 79: You need to add this env variable on the `serverless.yml` and `deploy.yml` files

gateways/token_balances_api_gateway.py
```diff
@@ -0,0 +1,111 @@
...
+
+class TokenBalancesApiGateway:
+    """ Gateway to interact with the Hathor ElasticSearch cluster with the token balances index
+    """
+
+    def __init__(
+        self,
+        elastic_search_client: Optional[Elasticsearch] = None,
+    ) -> None:
+        self.elastic_search_client = ElasticSearchClient(elastic_index=ELASTIC_TOKEN_BALANCES_INDEX,
```
lucas3003, line 26: Need to check if `elastic_search_client` was already passed as parameter of the function

gateways/token_balances_api_gateway.py
```diff
@@ -0,0 +1,111 @@
...
+                                                         client=elastic_search_client)
+
+    def get_token_information(self, token_id: str) -> dict:
+        """Retrieve total number of addresses and transactions for a given token_id
+        """
+
+        body = {
+            'size': 0,
```
lucas3003, line 34: Isn't this saying to return 0 objects?

gateways/token_balances_api_gateway.py
andreabadesso: This check is already done in `ElasticSearchClient`, `client` is optional, it will be created if it is not passed

This is the same behavior done in the token list gateway

handlers/token_balance_api.py
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+
+@ApiGateway()
+def get_token_information(
+    event: ApiGatewayEvent,
+    _context: LambdaContext,
+    token_balances_api: Optional[TokenBalancesApi] = None
+) -> dict:
+    """Get token information from a given token_id
```
lucas3003, line 50: I did not get if pure token information is get here or applied to the balance context.

serverless.yml
```diff
@@ -199,6 +199,50 @@ functions:
...
+    timeout: 6
+    maximumRetryAttempts: 0
+    package:
+      patterns:
+        - 'handlers/token_balance_api.py'
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: token_balances/information
```
lucas3003, line 237: maybe token_balances/{id} would be sufficient here? I do not know if the word `information` aggregate here as API name.

handlers/token_api.py
```diff
@@ -24,6 +24,9 @@ def get_tokens(
     search_after = event.query.get("search_after") or ""
     search_after_list = []
 
+    if sort_by == "uid":
+        sort_by = "id"
```
luislhl, line 28: This is to support requests from the Tokens balance screen as well, right? Should we add a comment explaining it?

handlers/token_balance_api.py
```diff
@@ -0,0 +1,63 @@
...
+
+@ApiGateway()
+def get_token_balances(
+    event: ApiGatewayEvent,
+    _context: LambdaContext,
+    token_balances_api: Optional[TokenBalancesApi] = None
+) -> dict:
+    """Get token balances from user search"""
+
+    token_balances_api = token_balances_api or TokenBalancesApi()
```
luislhl, line 19: This is equivalent to setting a default in the function parameters, right?

In case you think it's better:
```py
def get_token_balances(
    event: ApiGatewayEvent,
    _context: LambdaContext,
    token_balances_api: TokenBalancesApi = TokenBalancesApi()
) -> dict:
```

handlers/token_balance_api.py
```diff
@@ -0,0 +1,63 @@
...
+
+    token_id = event.query.get('token_id') or "00"
```
luislhl, line 21: Maybe we could have a constant to keep this `00`, like `HTR_TOKEN_ID = 00` (it could be in this same file)

Just to improve reading of the code

gateways/clients/elastic_search_client.py
```diff
@@ -19,7 +19,10 @@ def __init__(self, client: Optional[Elasticsearch]) -> None:
                 basic_auth=(ELASTIC_USER, ELASTIC_PASSWORD)
             )
 
-        self.utils = ElasticSearchUtils()
+        self.utils = ElasticSearchUtils(elastic_index=elastic_index)
+
+    def run(self, payload: dict) -> dict:
+        return dict(self.client.search(**payload))
 
     def make_query(self, search_text: str, sort_by: str, order: str, search_after: List[str]) -> dict:
```
luislhl, line 27: You weren't able to use this method because it's too specific for the other case we had before to list tokens, right?

Shouldn't we extract this logic of building the query from this class, to have the old case build its query outside as well?

Or is there any change this query will be reusable in other situations?

gateways/token_balances_api_gateway.py
```diff
@@ -0,0 +1,111 @@
...
+    'address': 'keyword',
+    'token_id': 'keyword',
+    'unlocked_balance': 'long',
+    'locked_balance': 'long',
+    'total': 'long',
+}
+
+
+class TokenBalancesApiGateway:
+    """ Gateway to interact with the Hathor ElasticSearch cluster with the token balances index
```
luislhl, line 19: I think adding a link to some documentation in Elasticsearch about how queries are built would be useful here for readers to easily learn more about what the code is doing.

common/configuration.py
andreabadesso: Thanks!

Added in 56b217f9118402cd0360e22f162795303b115324

gateways/clients/elastic_search_client.py
andreabadesso: I left it this way because every use case that has an index that has searchable fields will be able to reuse this query, even though we are not using it on the token balances query

handlers/token_api.py
andreabadesso: No, actually this was a design decision done on the token list API, I just moved this if outside of the `elastic_search_utils` method

The idea is to allow the user to sort by either `uid` or `id` (they are the same thing)

I removed it from `elastic_search_utils` `build_search_query`to be able to generalize it

handlers/token_balance_api.py
andreabadesso: I like your suggestion, refactored in 5ec74623639c35737c55020a1ac1e892927b7c53

handlers/token_balance_api.py
andreabadesso: We could get it from hathor lib, but we don't have it on the project

Using a constant on 2f185863d50dbca7e21d708208018ae48294b105

handlers/token_balance_api.py
andreabadesso: It is applied to the balance context, the information retrieved is the sum of all transactions on balances for a given token_id and the total number of addresses

That's why I called it `get_token_Information` in the context of the `token_balance_api`

serverless.yml
andreabadesso: I think `token_balances/{id}` feels like it is a token balance for a given token_id

The idea is to have some information about token_balances for a given token_id, that's why I used `token_balances`/`information`

gateways/token_balances_api_gateway.py
```diff
@@ -0,0 +1,111 @@
...
+        body = {
+            'size': int(ELASTIC_RESULTS_PER_PAGE) + 1,  # Last element is to check if there is next page.
+            'sort': [
+                primary_sort_key,
+                tie_break_sort_key
+            ],
+            'index': ELASTIC_TOKEN_BALANCES_INDEX,
+            'request_timeout': int(ELASTIC_SEARCH_TIMEOUT),
+            'query': {
+                'bool': {
```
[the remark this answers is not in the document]
andreabadesso: Refactored in 8983596c90de8e453b038218d0b1f5692046de50

Thanks!

luislhl (approved):

lucas3003 (approved):

PR #164, opened 2022-06-02
feat:add api to get last block available on elasticsearch

lucas3003:
### Acceptance Criteria

- A new endpoint on explorer-service is being created, where the block with the biggest height that is not voided is being returned to the client. We will use this information on front-end to inform the last sync from ElasticSearch.

This is an example of a response (Mocked data):

```
{
    "hits": [
        {
            "tx_id": "00a1786694f2b2248c4272e64f9f414759322b4e6d5e40d39cc5b5aedfd70dfb",
            "timestamp": "2022-05-09T18:55:47Z",
            "version": 0,
            "voided": false,
            "height": 1740645,
            "weight": 60.66999816894531
            "hash_rate": 123456
        }
    ],
    "has_next": false
}
```


serverless.yml
```diff
@@ -262,6 +263,26 @@ functions:
...
+    timeout: 15
+    maximumRetryAttempts: 0
+    package:
+      patterns:
+        - 'handlers/block_api.py'
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: blocks/biggest_height
```
r4mmer, line 277: Wouldn't something like `blocks/bestchain_height` or `blocks/current_height` be more explicit of what the api returns?

tests/unit/gateways/test_block_api_gateway.py
```diff
@@ -0,0 +1,80 @@
+from unittest.mock import MagicMock
+
+from pytest import fixture
+
+from gateways.block_api_gateway import BlockApiGateway
+
+ELASTIC_SEARCH_BIGGEST_HEIGHT_SUCCESSFUL_RAW_RESPONSE = {
```
r4mmer, line 7: I think we should have these big response bodies on the fixture folder, what do you think?

handlers/block_api.py
```diff
@@ -0,0 +1,27 @@
...
+
+@ApiGateway()
+def get_block_with_biggest_height(
+    event: ApiGatewayEvent,
+    _context: LambdaContext,
+    block_api: Optional[BlockApi] = None
+) -> dict:
+    """Get the block with biggest height available on the ElasticSearch"""
+
+    block_api = block_api or BlockApi()
```
luislhl, line 18: Should we just use a default in the function params instead?

Like this:
```
block_api: BlockApi = BlockApi()
```

gateways/block_api_gateway.py
```diff
@@ -0,0 +1,58 @@
...
+                    'height': {
+                        'order': 'desc'
+                    }
+                }
+            ],
+            'size': 1
+        }
+
+        elastic_search_utils = ElasticSearchUtils(elastic_index=ELASTIC_TX_INDEX)
+        return elastic_search_utils.treat_response(self.elastic_search_client.run(body))
```
luislhl, line 58: I think we are not treating errors that could happen in the elasticsearch client, right?

Like the ones in https://elasticsearch-py.readthedocs.io/en/v8.2.2/exceptions.html

Shouldn't we treat them to prevent the service from returning 500?

serverless.yml
```diff
@@ -262,6 +263,26 @@ functions:
...
+          method: get
+          cors: true
+          caching:
+            enabled: true
+            ttlInSeconds: 300
```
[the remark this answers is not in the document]
lucas3003: I will change to 60, because this is the interval ElasticSearch gets new data from Wallet Service

tests/unit/utils/elastic_search/transformations/test_tx.py
```diff
@@ -0,0 +1,42 @@
...
+        'weight': 60.66999816894531,
+        'updated_at': '2022-06-01T02:36:14Z',
+        'height': 1740645,
+        'created_at': '2022-06-01T02:31:49Z',
+        'version': 0,
+        'hash_rate': 123
+    }
+}
+
+ELASTIC_SEARCH_HIT_MAPPED = {
```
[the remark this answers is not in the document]
lucas3003: I will remove it in the next commit

gateways/block_api_gateway.py
```diff
@@ -0,0 +1,58 @@
+from typing import Optional
+
+from elasticsearch import Elasticsearch
+
+from common.configuration import ELASTIC_TX_INDEX
+from gateways.clients.elastic_search_client import ElasticSearchClient
+from utils.elastic_search.elastic_search_utils import ElasticSearchUtils
+
+BLOCK_TX_VERSION = 0
```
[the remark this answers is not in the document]
lucas3003: Done in the next commit

gateways/block_api_gateway.py
```diff
@@ -0,0 +1,58 @@
...
+        """
+
+        body = {
+            'index': ELASTIC_TX_INDEX,
+            'query': {
+                'bool': {
+                    'must': [
+                        {
+                            'match': {
+                                'version': BLOCK_TX_VERSION,
```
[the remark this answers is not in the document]
lucas3003: Done adding merged mining. I will check if it gets entry without version

serverless.yml
```diff
@@ -262,6 +263,26 @@ functions:
               - name: request.querystring.search_after
               - name: request.header.origin
 
+  get_block_with_biggest_height:
+    handler: handlers/block_api.get_block_with_biggest_height
+    timeout: 15
```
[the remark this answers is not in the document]
lucas3003: We have a constant for the ES Timeout, which is 25 seconds. I changed this to 30 seconds.

gateways/block_api_gateway.py
lucas3003: Done in last review. I've caught some of the exceptions from ElasticSearch and gave meaningful message to the client.

handlers/block_api.py
lucas3003: Done in the last review

tests/unit/usecases/test_block_api.py
```diff
@@ -0,0 +1,45 @@
...
+            'height': 1740645,
+            'weight': 60.66999816894531,
+            'hash_rate': 123
+        }
+    ],
+    'has_next': False
+}
+
+
+class TestBlockApi:
```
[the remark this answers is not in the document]
lucas3003: I've included a test case for a TransportError (5xx error)

tests/unit/gateways/test_block_api_gateway.py
lucas3003: Done in last review

serverless.yml
lucas3003: I agree. Changed to `blocks/bestchain_height`

gateways/block_api_gateway.py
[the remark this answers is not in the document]
lucas3003: I just checked that if we insert a document on ElasticSearch without a version, it will not be fetched on our query, as expected.

luislhl (approved):

tests/unit/utils/elastic_search/transformations/test_tx.py
```diff
@@ -0,0 +1,8 @@
+from tests.fixtures.elastic_search_fixtures import ELASTIC_SEARCH_RAW_HIT, ES_HIT_TO_RESULT_EXPECTED_RETURN
+from utils.elastic_search.transformations.tx import es_hit_to_result
+
+
+class TestTx:
+    def test_tx_transformation(self):
```
r4mmer, line 6: Is there no "error" case?
For instance if the es hit cannot be transformed because it's missing some property.

gateways/block_api_gateway.py
```diff
@@ -0,0 +1,80 @@
...
+
+        try:
+            elastic_search_response = self.elastic_search_client.run(body)
+            return elastic_search_utils.treat_response(elastic_search_response)
+        except RequestError:
+            return {'error': 'Bad request to ElasticSearch', 'status': 400}
+        except AuthorizationException:
+            return {'error': 'Explorer Service was not authorized to access ElasticSearch', 'status': 403}
+        except TransportError:
+            return {'error': 'An error ocurred before an HTTP response arrived',  'status': 500}
```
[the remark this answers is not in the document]
lucas3003: I agree! I will put it inside the usecases and make a better message

gateways/block_api_gateway.py
```diff
@@ -0,0 +1,80 @@
+from typing import Optional
+
+from elasticsearch import AuthorizationException, Elasticsearch, RequestError, TransportError
+from hathorlib.base_transaction import TxVersion
```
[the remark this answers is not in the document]
lucas3003: Will change in next commit

gateways/block_api_gateway.py
```diff
@@ -0,0 +1,80 @@
...
+                                                         )
+
+    def get_best_chain_height(self) -> dict:
+        """ Get the best chain height
+        """
+
+        body = {
+            'index': ELASTIC_TX_INDEX,
+            # This query can be transalated to
+            # ((TxVersion.REGULAR_BLOCK or TxVersion.MERGE_MINED_BLOCK) and voided == False)
```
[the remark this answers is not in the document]
lucas3003: I agree. This will make the query complete. I will update on the next commit

tests/unit/utils/elastic_search/transformations/test_tx.py
lucas3003: I will add a new `except`on `usecases/block_api` to inform client that document is missing information

usecases/block_api.py
```diff
@@ -0,0 +1,22 @@
...
+        try:
+            return self.block_api_gateway.get_best_chain_height()
+        except exceptions.RequestError:
+            return {'error': 'bad_request', 'status': 400}
+        except exceptions.AuthorizationException:
+            return {'error': 'not_authorized', 'status': 403}
+        except exceptions.TransportError:
+            return {'error': 'elasticsearch_transport_error',  'status': 500}
+        except KeyError:
+            return {'error': 'elasticsearch_document_missing_information',  'status': 500}
```
r4mmer, line 22:
```suggestion
        except exceptions.RequestError:
            logger.error('ElasticSearch request error')
            raise ApiError('invalid_parameters')
        except exceptions.AuthorizationException:
            logger.error('ElasticSearch authorization error')
            raise ApiError('not_authorized')
        except (exceptions.TransportError, KeyError):
            logger.error('ElasticSearch transport error')
            raise ApiError('internal_error')
```

The `logger` is just as an example, we could log the errors to make it easier to check the source of the error.
But I think we should use the `ApiError`, so that we have a unified way to return api errors and not have each function implement it's own way to handle api errors

r4mmer (changes requested):

usecases/block_api.py
lucas3003: Changed in the new commit. I've included the ApiError and the logger to enhance the traceability of the error

r4mmer (approved):

PR #174, opened 2022-06-20
feat: new function to update token metadata on s3

tuliomir:
This PR aims to build a new route with the functionality to update the token metadata on S3.

It is a dependency of HathorNetwork/hathor-wallet-service#235, which listens to new transactions, identifies NFT's and calls the `explorer-service` lambda to create/update the S3 metadata for them.

### Acceptance Criteria
- [X] Should have a new lambda `put_dag_metadata_handler` receiving the token's metadata as a JSON on the payload


serverless.yml
```diff
@@ -14,7 +14,7 @@ provider:
   region: ${env:AWS_DEFAULT_REGION}
   lambdaHashingVersion: 20201221
   logs:
-    restApi:
+    restApi: # TODO: add data to make the serverless create this role by itself
```
r4mmer, line 17: We should create an issue for this as well

serverless.yml
```diff
@@ -23,6 +23,10 @@ provider:
       - ${env:AWS_SUBNET_ID_1}
       - ${env:AWS_SUBNET_ID_2}
       - ${env:AWS_SUBNET_ID_3}
+  apiGateway:
+    apiKeys:
+      - name: updateS3Key
```
r4mmer, line 28: Add the stage on the apikey name.
This will make serverless create a different apikey for each stage (like described [here](https://www.serverless.com/framework/docs/providers/aws/events/apigateway#setting-api-keys-for-your-rest-api)).

And should the apikey be an `updateS3Key` or should we name it like an admin key?

serverless.yml
```diff
@@ -71,6 +75,7 @@ custom:
   s3:
     host: localhost
     directory: /tmp
+    buckets: hathor-explorer-metadata-dev
```
r4mmer, line 78: dev? Why not `hathor-explorer-metadata`?

r4mmer (changes requested):

serverless.yml
tuliomir: Opened #176 for this

serverless.yml
tuliomir: On a recent design change, this functionality will be used directly via lambda call, bypassing the `apiGateway`. So the keys were removed.

serverless.yml
tuliomir: This was changed on [e5ba6ac](https://github.com/HathorNetwork/hathor-explorer-service/pull/174/commits/e5ba6ac75a5a4fb833c567cc07543627583b3439) to use the `METADATA_BUCKET` environment variable, changing with the run context.

serverless.yml
```diff
@@ -284,10 +285,10 @@ functions:
               - name: request.header.origin
 
   get_dag_metadata_handler:
-    handler: handlers/get_dag_metadata_handler.handle
+    handler: handlers/dag_metadata_handler.handle_get
```
tuliomir, line 288: Note: the `get_dag_metadata_handler.py` file was renamed to `dag_metadata_handler.py` to accomodate both the `get` and `put` operations on S3.

usecases/metadata.py
```diff
@@ -17,3 +18,7 @@ def get(self, type: str, id: str) -> Optional[str]:
             return None
 
         return method(id)
+
+    def put_dag(self, type: str, id: str, content: dict) -> None:
```
pedroferreira1, line 22: Why do you need the `type` here?

gateways/metadata_gateway.py
```diff
@@ -65,3 +78,23 @@ def _metadata_bucket(self) -> str:
...
+        :param tokenUid: Token UID
+        :type tokenUid: str
+
+        :param metadata: Token metadata
+        :type metadata: Object
+
+        :return: S3 Upload metadata
+        :rtype: dict
+        """
+        response = self.s3_client.upload_file(
```
pedroferreira1, line 94: Is it possible to raise any errors here?

usecases/metadata.py
```diff
@@ -3,12 +3,13 @@
...
 
 
-class GetMetadata:
+class Metadata:
 
     def __init__(self, metadata_gateway: Optional[MetadataGateway] = None) -> None:
         self.metadata_gateway = metadata_gateway or MetadataGateway()
 
     def get(self, type: str, id: str) -> Optional[str]:
+        # Multiple metadatas can be obtained at once here
```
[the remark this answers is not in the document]
tuliomir: Comment removed on [abdb970](https://github.com/HathorNetwork/hathor-explorer-service/pull/174/commits/abdb970de4d304c04a700d45d4ea0ffa163b259c).

handlers/dag_metadata_handler.py
```diff
@@ -0,0 +1,55 @@
...
+    _context: LambdaContext,
+    metadata: Optional[Metadata] = None
+) -> dict:
+
+    metadata = metadata or Metadata()
+    if 'id' not in event.query:
+        raise ApiError('invalid_parameters')
+    hash = event.query['id']
+    content = event.body
+    metadata.put_dag('dag', hash, content)
```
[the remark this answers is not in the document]
pedroferreira1: I feel the method should be ready to receive a new content for metadata but assuming some data could already be saved in the S3, so we should fetch the data there and merge with the new content. Something like a `create_or_update_metadata` method. What do you think?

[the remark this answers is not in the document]
tuliomir: I think this will make this method safer and easier to use. I'll impement it :+1:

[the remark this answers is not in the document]
tuliomir: Implemented on [179967a](https://github.com/HathorNetwork/hathor-explorer-service/pull/174/commits/179967a77b93bbcf9e4ae0a4783850c511fefbc2).

pedroferreira1 (approved):

PR #178, opened 2022-07-21
feat: big addresses api

r4mmer:
# Acceptance criteria

- Create `/address/token` API: search the tokens of any address
- Create `/address/history` API: search the history of any address/token pair
- Create `/address/balance` API: query the balance of any address/token pair

# Description

The new APIs would allow the explorer to query for the address information (tokens, balance and history) without burdening the full-node and much faster since the data required is already indexed on the database.


.env.example
```diff
@@ -25,4 +25,8 @@ ELASTIC_PASSWORD=elastic_password
 ELASTIC_INDEX=elastic_index
 ELASTIC_TX_INDEX=elastic_tx_index
 ELASTIC_RESULTS_PER_PAGE=10
-ELASTIC_SEARCH_TIMEOUT=25
\ No newline at end of file
+ELASTIC_SEARCH_TIMEOUT=25
+WALLET_SERVICE_USERNAME=wallet_service_user
+WALLET_SERVICE_PASSWORD=password
```
lucas3003, line 30: I would avoid accessing directly wallet service's database, "surpassing" the service. Would it be feasible to implement an API on wallet-service to proxy this request to the database?

.env.example
```diff
@@ -25,4 +25,8 @@ ELASTIC_PASSWORD=elastic_password
 ELASTIC_INDEX=elastic_index
 ELASTIC_TX_INDEX=elastic_tx_index
 ELASTIC_RESULTS_PER_PAGE=10
-ELASTIC_SEARCH_TIMEOUT=25
\ No newline at end of file
+ELASTIC_SEARCH_TIMEOUT=25
+WALLET_SERVICE_USERNAME=wallet_service_user
```
lucas3003, line 29: minor: I would change this to `WALLET_SERVICE_DB_USERNAME` and `WALLET_SERVICE_PASSWORD` to `WALLET_SERVICE_DB_PASSWORD`

domain/wallet_service.py
```diff
@@ -0,0 +1,128 @@
...
+    :param height: Transaction height
+    :type height: int
+    """
+
+    tx_id: str
+    token_id: str
+    timestamp: int
+    balance: int
+    version: int
+    height: Union[int, None]
```
lucas3003, line 76: Is this equivalent of `height: Optional[int]`

domain/wallet_service.py
```diff
@@ -0,0 +1,128 @@
...
+    :type timestamp: int
+
+    :param balance: Transaction balance for the token
+    :type balance: int
+
+    :param version: Transaction version
+    :type type: int
+
+    :param height: Transaction height
+    :type height: int
```
lucas3003, line 68: Not matching with the variable

domain/wallet_service.py
```diff
@@ -0,0 +1,128 @@
...
+    :param token_id: Token unique id
+    :type token_id: str
+
+    :param name: Token name
+    :type name: str
+
+    :param symbol: Token symbol
+    :type symbol: str
+
+    :param createdAt: Timestamp this entry was created
```
lucas3003, line 104: Parameter not created

domain/wallet_service.py
```diff
@@ -0,0 +1,128 @@
...
+    :type createdAt: int
+
+    :param updatedAt: Last update timestamp
```
lucas3003, line 107: Parameter not created

handlers/wallet_service.py
```diff
@@ -0,0 +1,118 @@
...
+from common.errors import ApiError, RdsNotFoundError
+from usecases.wallet_service import WalletService
+from utils.wrappers.aws.api_gateway import ApiGateway, ApiGatewayEvent
+
+
+@ApiGateway()
+def handle_address_balance(
+    event: ApiGatewayEvent,
+    __: LambdaContext,
+    wallet_service: Union[WalletService, None] = None
```
lucas3003, line 15: Why we dont use `Optional` here?

handlers/wallet_service.py
```diff
@@ -0,0 +1,118 @@
...
+    address: Optional[str] = event.query.get("address")
+    token: Optional[str] = event.query.get("token")
+
+    if address is None or token is None:
+        raise ApiError("invalid_parameters")
+
+    response: Optional[dict]
+    try:
+        response = wallet_service.address_balance(address, token)
+    except RdsNotFoundError:
```
lucas3003, line 29: I saw that we also raise `RdsError('only one row expected')`. Should we also translate that to an `ApiError`?

handlers/wallet_service.py
```diff
@@ -0,0 +1,118 @@
...
+
+    limit: int
+    offset: int
+    try:
+        limit = int(limit_str)
+        offset = int(offset_str)
+    except ValueError:
+        raise ApiError("invalid_parameters")
+
+    if limit > 100:
```
lucas3003, line 69: Do we make explicit anywhere to the user that the `limit` is 100?

serverless.yml
```diff
@@ -566,3 +570,67 @@ functions:
...
+          method: get
+          cors: true
+          caching:
+            enabled: true
+            ttlInSeconds: 5
+            cacheKeyParameters:
+              - name: request.querystring.address
+              - name: request.querystring.token
+              - name: request.querystring.limit
+              - name: request.querystring.skip
```
lucas3003, line 615: I think `offset` should be here instead of `skip`

serverless.yml
```diff
@@ -566,3 +570,67 @@ functions:
...
+      - http:
+          path: address/tokens
+          method: get
+          cors: true
+          caching:
+            enabled: true
+            ttlInSeconds: 10
+            cacheKeyParameters:
+              - name: request.querystring.address
+              - name: request.header.origin
```
lucas3003, line 636: Should we also have `limit` and `skip` params here?

handlers/wallet_service.py
lucas3003: Oh, I saw that it is explicit on ApiDocs. You can ignore this comment

.env.example
r4mmer: Done, also changed `WALLET_SERVICE_HOST` to `WALLET_SERVICE_DB_HOST` and `WALLET_SERVICE_DB` to `WALLET_SERVICE_DB_NAME`

domain/wallet_service.py
r4mmer: Yes, `Optional[X]` is an alias to `Union[X, None]`.

domain/wallet_service.py
r4mmer: Fixed. The correct one is the variable since height will be `None` for mempool transactions.

domain/wallet_service.py
r4mmer: Removed.

domain/wallet_service.py
r4mmer: Removed.

handlers/wallet_service.py
r4mmer: `Optional[X]` and `Union[X, None]` are the same, so they can be used interchangeably but usually I prefer `Union[X, None]` for function args with default and properties.
This is not completely arbitrary, usually typing a parameter `Optional` gives the impression that if they are not present it will be `None` automatically, but you still have to add the `= None` after for the default value because `Optional` is just an alias.

That's why you will see more `Union` on class and function definitions and `Optional` on the function body.
But I'm not against just changing all to `Optional` since they are interchangeable. What do you think?

handlers/wallet_service.py
r4mmer: Since the query itself limits the response to 1 row it should never happen, and if it does it is an error that should be alerted.
Letting it raise the exception and returning an 500 seems like the correct option in this case.

handlers/wallet_service.py
r4mmer: Only on the documentation.

serverless.yml
r4mmer: Fixed

serverless.yml
r4mmer: Added

.env.example
r4mmer: We will be using a read replica and not connecting to the wallet service db directly, so it will not interfere with the wallet-service in any capacity.

text/api-docs.yml
```diff
@@ -0,0 +1,16 @@
...
+servers:
+  - url: "http://localhost:3001/dev/"
+  - url: "https://explorer-service.hathor.network/"
+paths:
+  /wallet/address/tokens:
+    $ref: "./openapi/wallet_service.yml#/methods/address_tokens"
+  /wallet/address/balance:
+    $ref: "./openapi/wallet_service.yml#/methods/address_balance"
+  /wallet/address/history:
+    $ref: "./openapi/wallet_service.yml#/methods/address_history"
```
andreabadesso, line 16: Are those paths correct or should we remove `wallet/`?

andreabadesso (approved):

handlers/wallet_service.py
lucas3003: Thank you for the explanation! I agree with maintaing `Union` here.

lucas3003 (approved):

PR #188, opened 2022-08-12
feat: always include htr

r4mmer:
### Acceptance criteria

- Include HTR on the first page of the address tokens API if it exists on the address history

### DIFF
--- a/gateways/clients/wallet_service_db_client.py
+++ b/gateways/clients/wallet_service_db_client.py
@@ -48,10 +48,15 @@ SELECT address_balance.token_id AS token_id,
        token.name AS name,
        token.symbol AS symbol,
        COUNT(*) OVER() as total
-FROM token RIGHT OUTER JOIN address_balance ON token.id = address_balance.token_id
+FROM token INNER JOIN address_balance ON token.id = address_balance.token_id
 WHERE address_balance.address = :address
 ORDER BY address_balance.updated_at DESC
 LIMIT :limit OFFSET :offset'''
+
+address_has_htr_query: str = '''\
+SELECT '00' as token_id, 'Hathor' as name, 'HTR' as symbol
+WHERE EXISTS (SELECT 1 FROM address_balance WHERE address = :address and token_id = '00')
+'''
 
 
 def get_engine():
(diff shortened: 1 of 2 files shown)


gateways/clients/wallet_service_db_client.py
```diff
@@ -48,11 +48,16 @@
        token.name AS name,
        token.symbol AS symbol,
        COUNT(*) OVER() as total
-FROM token RIGHT OUTER JOIN address_balance ON token.id = address_balance.token_id
+FROM token INNER JOIN address_balance ON token.id = address_balance.token_id
```
r4mmer, line 51: This was changed to an `INNER JOIN` because it will not include HTR.
Since we already search HTR outside of this query.

tests/unit/gateways/clients/test_wallet_service_db_client.py
```diff
@@ -123,3 +102,90 @@ def test_get_address_balance_errors(self, engine, connection):
...
+            if q.text == address_has_htr_query:
+                cursor = MagicMock()
+                cursor.one_or_none.return_value = self.row_from_dict('htr_dict_value')
+                return cursor
+            else:
+                return [
+                    self.row_from_dict('dict_value_1'),
+                    self.row_from_dict('dict_value_2'),
+                ]
+        connection.execute.side_effect = cursor_func
```
r4mmer, line 140: We use `side_effect` here because we need different return values depending on which query is being executed.

gateways/clients/wallet_service_db_client.py
r4mmer: After https://github.com/HathorNetwork/hathor-wallet-service/pull/273 is merged this will also include HTR.

Should we explicitly exclude HTR? As in add a `WHERE token.id != '00'`

pedroferreira1 (approved):

andreabadesso (approved):

gateways/clients/wallet_service_db_client.py
andreabadesso: I think that should work

PR #189, opened 2022-08-15
feat: get token by token_id API and add transactions count to token return

andreabadesso:
### Acceptance Criteria
- We should have an API to retrieve a specific token by its token_id
- We should add the `transactions_count` attribute to the token response
- We should remove the `transactions` aggregation from the token information API

### Security Checklist
- [X] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

### DIFF
--- a/gateways/token_balances_api_gateway.py
+++ b/gateways/token_balances_api_gateway.py
@@ -39,7 +39,6 @@ class TokenBalancesApiGateway:
     def get_token_information(self, token_id: str) -> dict:
         """Retrieve total number of addresses and transactions for a given token_id
         """
-
         body = {
             'size': 0,
             'query': self._build_filtered_query(token_id),
@@ -51,18 +50,12 @@ class TokenBalancesApiGateway:
                         'field': 'address.keyword'
                     }
                 },
-                'transaction_sum': {
-                    'sum': {
-                        'field': 'transactions'
-                    }
-                }
             }
         }
 
         response = self.elastic_search_client.run(body)
 
         return {
-            'transactions': response['aggregations']['transaction_sum']['value'],
             'addresses': response['aggregations']['address_count']['value']
         }
 
--- a/handlers/token_api.py
+++ b/handlers/token_api.py
@@ -58,7 +58,7 @@ def get_token(
     response = token_api.get_token(token_id)
 
     if len(response["hits"]) == 0:
-        raise ApiError("Token not found")
+        raise ApiError("not_found")
 
     return {
         "statusCode": 200,
--- a/handlers/token_api.py
+++ b/handlers/token_api.py
@@ -57,7 +57,7 @@ def get_token(
 
     response = token_api.get_token(token_id)
 
-    if len(response['hits']) == 0:
+    if len(response["hits"]) == 0:
         raise ApiError("Token not found")
 
     return {
--- a/handlers/token_api.py
+++ b/handlers/token_api.py
@@ -47,8 +47,9 @@ def get_token(
     _context: LambdaContext,
     token_api: TokenApi = TokenApi()
 ) -> dict:
-    """Get a specific token from its token_id"""
-
+    """Get a specific token from the elastic_search index
+       given an token_id
+    """
     token_id = event.path["token_id"]
 
     if token_id is None:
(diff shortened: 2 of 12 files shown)


gateways/token_balances_api_gateway.py
```diff
@@ -51,18 +50,12 @@ def get_token_information(self, token_id: str) -> dict:
                         'field': 'address.keyword'
                     }
                 },
-                'transaction_sum': {
```
andreabadesso, line 54: This is not only not needed as this information will be gathered by using the token index but it is invalid, as the aggregation is being limited by the max aggregation size from elastic search (limited at 1000 rows)

serverless.yml
```diff
@@ -268,6 +268,31 @@ functions:
...
+    events:
+      - http:
+          path: tokens/{token_id}
+          method: get
+          cors: true
+          caching:
+            enabled: true
+            ttlInSeconds: 300
+            cacheKeyParameters:
+              - name: request.querystring.search_text
```
pedroferreira1, line 290: Do we need these? I thought they were used only by the `get_tokens` API, not the get one token.

handlers/token_api.py
```diff
@@ -39,3 +39,31 @@ def get_tokens(
...
+       given an token_id
+    """
+    token_id = event.path["token_id"]
+
+    if token_id is None:
+        raise ApiError("Invalid token_id")
+
+    response = token_api.get_token(token_id)
+
+    if len(response['hits']) == 0:
```
pedroferreira1, line 60: Single quotes with double quotes

handlers/token_api.py
```diff
@@ -39,3 +39,31 @@ def get_tokens(
...
+        raise ApiError("Token not found")
+
+    return {
+        "statusCode": 200,
+        "body": json.dumps(response or {}),
```
pedroferreira1, line 65: Why do you need this `or`? If there's a possibility of `response` being `None`, then the expression above `if len(response['hits']) == 0:` would raise an error, right?

tests/unit/usecases/test_token_api.py
```diff
@@ -48,3 +50,32 @@ def test_get_tokens(self, token_api_gateway):
...
+                    "nft": False
+                }
+            ],
+        }
+
+        token_api_gateway.get_token = MagicMock(return_value=obj)
+
+        token_api = TokenApi(token_api_gateway)
+
+        result = token_api.get_token(
```
pedroferreira1, line 76: Should we test the API with no parameters, wrong token id, ... ?

serverless.yml
```diff
@@ -268,6 +268,31 @@ functions:
               - name: request.querystring.search_after
               - name: request.header.origin
 
+  # This get_token will get tokens from the elasticsearch index
+  es_get_token_handler:
```
r4mmer, line 272: I know this project style is to use snake case for lambda names but AWS translates to camel case replacing each underscore with `Underscore` for some internal names (e.g. policy statement name) this makes the name longer and easier to hit the limit, failing the deploy.

I think we should create new lambdas with camel case to avoid these issues.

tests/unit/usecases/test_token_api.py
andreabadesso: Added tests for the gateway on 31fb9ac052a7ef10042052359a9fb8e61d795632

A call with no parameters would fall on the token list API as the only difference between them is the path parameter `token_id`

handlers/token_api.py
```diff
@@ -39,3 +39,31 @@ def get_tokens(
...
+    """
+    token_id = event.path["token_id"]
+
+    if token_id is None:
+        raise ApiError("Invalid token_id")
+
+    response = token_api.get_token(token_id)
+
+    if len(response['hits']) == 0:
+        raise ApiError("Token not found")
```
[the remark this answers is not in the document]
andreabadesso: Yes we should, thanks!

Done in d45ebd0253aba75c5b8c3794ef6be2be86c52f28

handlers/token_api.py
andreabadesso: Fixed

serverless.yml
andreabadesso: Removed the unused `cacheKeyParameters`

r4mmer (approved):

pedroferreira1 (approved):

PR #191, opened 2022-08-18
fix: deploy config for new envs

r4mmer:
# Acceptance criteria

Allow configuring both testnet and mainnet deploys on github

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -91,6 +91,10 @@ jobs:
           ELASTIC_INDEX: testnet-token
           ELASTIC_TX_INDEX: testnet-tx
           ELASTIC_TOKEN_BALANCES_INDEX: testnet-token-balance
+          WALLET_SERVICE_DB_USERNAME: ${{ secrets.TESTNET_WALLET_SVC_DB_USERNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -91,6 +91,10 @@ jobs:
           ELASTIC_INDEX: testnet-token
           ELASTIC_TX_INDEX: testnet-tx
           ELASTIC_TOKEN_BALANCES_INDEX: testnet-token-balance
+          WALLET_SERVICE_DB_USERNAME: ${{ secrets.TESTNET_WALLET_SVC_DB_USERNAME }}
```
luislhl, line 94: We could use [environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) instead of naming the secrets with a prefix `TESTNET_` or `MAINNET_`

environment support is already set up in this workflow (see lines 30 and 45), so you could just use `${{ secrets.WALLET_SERVICE_DB_USERNAME` }}, and when configuring the secret in the repo, make sure to configure each secret in their respective environment.

I suspect you could also add these new variables to the global env in line 8, instead of adding them here.

r4mmer: Done

tuliomir (approved):

luislhl (approved):

PR #193, opened 2022-08-23
fix: count tokens query

r4mmer:
### Acceptance Criteria
Since HTR is searched before any token, both queries we make will not contain all tokens so we should query for the total number of tokens separately.

### Security Checklist

### DIFF
--- a/gateways/clients/wallet_service_db_client.py
+++ b/gateways/clients/wallet_service_db_client.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, List, Optional
+from typing import TYPE_CHECKING, List, Optional, Tuple
 
 from sqlalchemy.engine import URL, create_engine
 from sqlalchemy.exc import MultipleResultsFound, NoResultFound
@@ -47,7 +47,6 @@ address_tokens_query: str = '''\
 SELECT address_balance.token_id AS token_id,
        token.name AS name,
        token.symbol AS symbol,
-       COUNT(*) OVER() as total
 FROM token INNER JOIN address_balance ON token.id = address_balance.token_id
 WHERE address_balance.address = :address AND token.id != '00'
 ORDER BY address_balance.updated_at DESC
@@ -57,6 +56,12 @@ address_has_htr_query: str = '''\
 SELECT '00' as token_id, 'Hathor' as name, 'HTR' as symbol
 WHERE EXISTS (SELECT 1 FROM address_balance WHERE address = :address and token_id = '00')
 '''
+
+address_tokens_count_query = '''\
+SELECT COUNT(*) as total
+FROM token INNER JOIN address_balance
+ON token.id = address_balance.token_id
+WHERE address_balance.address = :address'''
 
 
 def get_engine():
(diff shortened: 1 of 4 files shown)


gateways/clients/wallet_service_db_client.py
```diff
@@ -58,6 +57,12 @@
 WHERE EXISTS (SELECT 1 FROM address_balance WHERE address = :address and token_id = '00')
 '''
 
+address_tokens_count_query = '''\
+SELECT COUNT(*) as total
+FROM token INNER JOIN address_balance
+ON token.id = address_balance.token_id
+WHERE address_balance.address = :address'''
```
r4mmer, line 64: I could query only on `address_balance` since we only need to count distincts `token_id` for an address.
I prefer querying with the `inner join` to make sure the response here will always return the same number of tokens that can be found with the `address_tokens_query` (plus HTR).

The address_balance and token tables should never be "out of sync" but I had this issue on my local test (which is not fully synced).

gateways/clients/wallet_service_db_client.py
```diff
@@ -105,10 +110,19 @@ def get_address_history(self, address: str, token: str, limit: int, offset: int)
 
         return result
 
-    def get_address_tokens(self, address: str, limit: int, offset: int) -> List[dict]:
+    def get_address_tokens(self, address: str, limit: int, offset: int) -> Tuple[int, List[dict]]:
```
luislhl, line 113: Could you add a short description of what this function does?

I think it could make reading the code easier for someone that isn't used to it.

gateways/clients/wallet_service_db_client.py
```diff
@@ -105,10 +110,19 @@ def get_address_history(self, address: str, token: str, limit: int, offset: int)
...
         result: List[dict] = []
         found_htr: bool = False
+        total: int = 0
         with self.engine.connect() as connection:
+            cursor = connection.execute(text(address_tokens_count_query), address=address)
```
luislhl, line 118: Does SQL Alchemy protect us against SQL Injection when executing a raw query like this?

r4mmer: SQLAlchemy quotes the parameters (escaping if needed).
The most recent version which is not fully protected against SQL Injection is 1.2.18 (more on this [here](https://security.snyk.io/vuln/SNYK-PYTHON-SQLALCHEMY-173678)) and we use 1.4.39.

gateways/clients/wallet_service_db_client.py
r4mmer: Added docstrings explaining the method logic and some comments

gateways/clients/wallet_service_db_client.py
luislhl: Great

gateways/clients/wallet_service_db_client.py
luislhl: Thanks

luislhl (approved):

andreabadesso (approved):

PR #218, opened 2022-10-25
chore: improve logging

luislhl:
### Acceptance Criteria
- We should log some events in the DataCollector daemon so we know it's still working just by looking at the logs.

### Security Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ install:
 build:
 	docker-compose build
 
+.ONESHELL:
 .PHONY: deploy-daemons
 deploy-daemons:
 	if [ -z "${AWS_ACCOUNT_ID}" ]; then \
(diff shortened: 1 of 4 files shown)


andreabadesso (approved):

Makefile
```diff
@@ -71,6 +71,7 @@ install:
 build:
 	docker-compose build
 
+.ONESHELL:
```
r4mmer, line 74: Any specific reason for this?
Was there a bug due to the multiline commands?

r4mmer (approved):

Makefile
luislhl: Yes, the env var exported with `export DOCKER_IMAGE_TAG="dev-$$commit-$$timestamp";` was not available in subsequent commands

Issue #133, opened 2022-02-25
Timeout on the Lambdas leave no log about the problem


luislhl:
We got some timeouts from the Lambda `node_api_get_transaction`. When a timeout happens in the Lambda, we have no way to know what exactly happened.

To improve this, we should configure a timeout in the client that is lesser than the Lambda timeout, so it times out earlier and logs something about the error.

alexruzenhack:
The AWS Lambda Function timeout is configured by default with 6 seconds in Serverless. That said, the client should have timeout of 5 seconds.

PR #251, opened 2023-07-02
chore: set timeout for node api calls

alexruzenhack:
### Acceptance Criteria
- turn explicit the timeout value for Node Api lambdas
- set timeout for Node Api client calls

### Security Checklist

Closes: https://github.com/HathorNetwork/hathor-explorer-service/issues/133

### DIFF
--- a/gateways/node_api_gateway.py
+++ b/gateways/node_api_gateway.py
@@ -16,7 +16,9 @@ from gateways.clients.hathor_core_client import (
     HathorCoreClient,
 )
 
-# The default lambda timeout for Node API Lambdas is set to 6 seconds.
+# The default lambda timeout for Node API Lambdas is set to
+# 6 seconds. Therefore, the client should have a lower timeout,
+# to let the lambda shut down gracefully.
 NODE_API_TIMEOUT_IN_SECONDS = 5
 
 
@@ -59,7 +61,9 @@ class NodeApiGateway:
         return self.hathor_core_client.get(
             ADDRESS_BALANCE_ENDPOINT,
             params={"address": address},
-            timeout=14,  # lambda timeout 15s
+            # lambda timeout 15s, which is the sum of all lambda's steps timeout
+            # see: https://github.com/HathorNetwork/hathor-explorer-service/pull/93
+            timeout=10,
         )
 
     def get_address_search(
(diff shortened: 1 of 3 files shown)


gateways/node_api_gateway.py
```diff
@@ -16,6 +16,9 @@
     HathorCoreClient,
 )
 
+# The default lambda timeout for Node API Lambdas is set to 6 seconds.
```
tuliomir, line 19: I suggest informing on this comment that the timeouts set here will be shorter than the lambdas timeout itself.
Without navigating to the issue that originated this PR, it's not obvious why there is this difference.

gateways/node_api_gateway.py
```diff
@@ -122,21 +133,31 @@ def get_token_history(
             params["page"] = page
             params["timestamp"] = timestamp
 
-        return self.hathor_core_client.get(TOKEN_HISTORY_ENDPOINT, params=params)
+        return self.hathor_core_client.get(
+            TOKEN_HISTORY_ENDPOINT, params=params, timeout=NODE_API_TIMEOUT_IN_SECONDS
```
tuliomir, line 137: Why don't we refactor this broken down to three lines, as the other functions?

serverless.yml
```diff
@@ -390,6 +390,7 @@ functions:
 
   node_api_get_version:
     handler: handlers/node_api.get_version
+    timeout: 6
```
tuliomir, line 393: Maybe add a comment indicating this is informed in seconds?

tuliomir (approved):

gateways/node_api_gateway.py
```diff
@@ -56,7 +59,7 @@ def get_address_balance(self, address: str) -> Optional[dict]:
         return self.hathor_core_client.get(
             ADDRESS_BALANCE_ENDPOINT,
             params={"address": address},
-            timeout=10,
+            timeout=14,  # lambda timeout 15s
```
luislhl, line 62: You need to be careful because some Lambdas make more than one request to the outside. This is one of them.

See https://github.com/HathorNetwork/hathor-explorer-service/pull/93, which is the PR that set the timeout as 10 here.

Apparently it should remain as 10 if the Lambda timeout is 15.

gateways/node_api_gateway.py
```diff
@@ -86,25 +89,33 @@ def get_address_search(
         return self.hathor_core_client.get(
             ADDRESS_SEARCH_ENDPOINT,
             params=params,
-            timeout=10,
+            timeout=14,  # lambda timeout 15s
```
luislhl, line 92: This one is also part of a Lambda that makes other requests besides this.

I think it should remain as 10s.

luislhl (changes requested):
See my previous comments

serverless.yml
alexruzenhack: I will let it unchanged, because if I add this comment here, I would have to add this comment in all other place.

gateways/node_api_gateway.py
alexruzenhack: Oh! This is because of `black` code style.

luislhl (approved):

PR #256, opened 2023-07-20
feat: change push_tx to post

r4mmer:
### Acceptance Criteria
- Change `push_tx` to post

### Security Checklist

### DIFF
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -93,6 +93,7 @@ class HathorCoreAsyncClient:
             self.log.error("hathor_core_error", path=path, error=repr(e))
             callback({"error": repr(e)})
 
+
 class HathorCoreClient:
     """Client to make requests
 
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -157,7 +157,7 @@ class HathorCoreClient:
         url = parse.urljoin(f"https://{self.domain}", path)
 
         try:
-            response = requests.get(url, json=body, **kwargs)
+            response = requests.post(url, json=body, **kwargs)
             if response.status_code != 200:
                 self.log.warning(
                     "hathor_core_error",
@@ -212,7 +212,6 @@ class HathorCoreClient:
         :return: request response
         :rtype: Optional[dict]
         """
-        self.log.warning("hathor_post_request", path=path)
         text = self.post_text(path, body, **kwargs)
         if not text:
             return None
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -64,6 +64,34 @@ class HathorCoreAsyncClient:
             self.log.error("hathor_core_error", path=path, error=repr(e))
             callback({"error": repr(e)})
 
+    async def post(
+        self, path: str, callback: Callable[[dict], None], body: Optional[dict] = None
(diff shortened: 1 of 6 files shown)


tests/unit/usecases/test_token_balances_api.py
```diff
@@ -0,0 +1,42 @@
+import json
```
r4mmer, line 1: I created this test to match the coverage so it is unrelated to the core matter of the PR.
I chose the token balances api usecase since it had 0% coverage, now it has 100%.

gateways/clients/hathor_core_client.py
```diff
@@ -64,6 +64,35 @@ async def get(
             self.log.error("hathor_core_error", path=path, error=repr(e))
             callback({"error": repr(e)})
 
+    async def post(
```
r4mmer, line 67: While it is not used by push_tx I think we should support the same methods on both async and sync clients so I created the async post as well

r4mmer:
To test this api you need a valid signed and mined transaction hex.
This script helps with this:

```js
const hathorLib = require('@hathor/wallet-lib'); // or require('.') if running on wallet-lib

const connection = new hathorLib.Connection({
    network: 'mainnet',
    servers: ['https://node1.mainnet.hathor.network/v1a/'],
    connectionTimeout: 5000,
});

connection.on('wallet-load-partial-update', console.log)

const seed = '<put your seed here>';

const store = new hathorLib.MemoryStore();
const storage = new hathorLib.Storage(store);
const walletConfig = {
    seed,
    pinCode: '1234',
    password: '1234',
    connection,
    storage,
};

const wallet = new hathorLib.HathorWallet(walletConfig);


async function waitReady(w) {
  while (!w.isReady()) {
    await new Promise(res => setTimeout(() => { res() }, 1000));
  }
  console.log('wallet ready');
}

async function main() {
  wallet.start().then(console.log);
  await waitReady(wallet);
  const addr = await wallet.getCurrentAddress();

  const balance = await wallet.getBalance('00');
  console.log(JSON.stringify(balance, null, 2));

  const sendTx = new hathorLib.SendTransaction({
    pin: '1234',
    storage: wallet.storage,
    outputs: [{ type: 'p2pkh', address: addr.address, value: 1, token: '00' }],
  });

  const tx = await sendTx.run('mine-tx');
  // const tx = await sendTx.run('prepare-tx');
  console.log(tx);
  console.log(tx.toHex());
}

main();
```

tests/unit/usecases/test_token_balances_api.py
alexruzenhack, line 1: 🚲 this import is not being used. Can you remove?

alexruzenhack (approved):

tuliomir (commented):
The code looks good to me. :heavy_check_mark: 

I'm only wondering why we don't have refactored tests for the changed route.
Shouldn't the previous tests stop working as we changed the verb and the parameters from `query` to `body`?

r4mmer:
> The code looks good to me. ✔️
> 
> I'm only wondering why we don't have refactored tests for the changed route. Shouldn't the previous tests stop working as we changed the verb and the parameters from `query` to `body`?

The tests do not cover the api handlers, only the gateways and usecases, these receive hex_tx as a parameters and not the request dict.
So the tests stay the same while the handler just extracts the data from a different place in the request.

tuliomir (approved):

PR #260, opened 2023-08-02
chore: upgrade serverless version

r4mmer:
### Description

Serverless creates a lambda as a custom resource to handle logs, this lambda was running with node12 that has been deprecated and AWS does not allow it anymore.
The solution for this has been made in serverless v3 but a backport to v2 has been made ([reference](https://github.com/serverless/serverless/issues/12054#issuecomment-1632634253)).

The best long term solution would require upgrading to serverless v3, but we should check the changes in the serverless file syntax and other improvements on v3 before this upgrade.

### Acceptance Criteria
- Upgrade serverless to v2.72.4

### Observations

During the sandbox deployment I found some issues:
- Python 3.11 has some issues installing aiohttp, so I had to lock the python version on 3.9 (which is configured to run on the lambda)
- I used node v16.19.1 with npm version 8.19.3 and had to update the package-lock file, maybe we should try to enforce using node v18 which is recommended by aws.
- There is no sandbox deployment guides, even my sandbox deployment did not come fully configured since the explorer-service depends on some external resources, maybe we should create one

### Security Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,22 +1,9857 @@
 {
   "name": "hathor-explorer-service",
   "version": "0.8.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
-  "dependencies": {
-    "2-thenable": {
+  "packages": {
+    "": {
+      "name": "hathor-explorer-service",
+      "version": "0.8.0",
+      "license": "MIT",
+      "dependencies": {
+        "@apidevtools/swagger-cli": "^4.0.4",
+        "serverless": "^2.72.4",
+        "serverless-api-gateway-caching": "^1.7.2",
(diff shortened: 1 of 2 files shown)


luislhl (approved):

luislhl:
> There is no sandbox deployment guides, even my sandbox deployment did not come fully configured since the explorer-service depends on some external resources, maybe we should create one

Or we could try to make serverless create all dependencies, unless this is unfeasible. Should we have an issue for this?

> I used node v16.19.1 with npm version 8.19.3 and had to update the package-lock file, maybe we should try to enforce using node v18 which is recommended by aws.

I like the idea of enforcing a node version, to avoid the constant changes to package-lock.json when devs use different versions.

PR #276, opened 2023-10-26
feat: data collector health ping

luislhl:
### Acceptance Criteria

This feature will allow sending pings every time the data collector runs to an external uptime monitor service (see https://github.com/topics/uptime-monitor and https://github.com/topics/uptime for options, some of them are able to receive these pings)

There should be a feature-flag for this that is disabled by default.

The URL to send the ping to and the interval should be configurable, besides the api-key, if needed.

### Security Checklist

### DIFF
--- a/utils/healthcheck/healthcheck_ping_manager.py
+++ b/utils/healthcheck/healthcheck_ping_manager.py
@@ -1,0 +1,61 @@
+from datetime import datetime
+
+import aiohttp
+
+from common.configuration import (
+    HEALTHCHECK_DATA_COLLECTOR_URL,
+    HEALTHCHECK_PING_ENABLED,
+    HEALTHCHECK_PING_INTERVAL,
+    HEALTHCHECK_SERVICE_API_KEY,
+)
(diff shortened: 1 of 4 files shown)


andreabadesso (approved):

utils/healthcheck/healthcheck_ping_manager.py
```diff
@@ -0,0 +1,61 @@
...
+)
+from common.logging import get_logger
+
+logger = get_logger()
+
+
+class HealthcheckPingManager:
+    def __init__(self) -> None:
+        self.targets = {
+            "data_collector": {
```
r4mmer, line 19: Do we expect any expansion on this healthcheck manager?

r4mmer (approved):

utils/healthcheck/healthcheck_ping_manager.py
luislhl: I just preferred to create it in a expansible way, but I don't expect this to be needed in the short-term.

PR #280, opened 2023-11-01
chore:  allows proteina staging URLs on CORS rules for testnet and mainnet environments

alexruzenhack:
### Acceptance Criteria
- Add CORS configuration for proteina stating environment of the site on our dev-testnet environment

### Security Checklist

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -2,7 +2,7 @@ name: deploy
 
 on:
   push:
-    branches: [main, dev]
+    branches: [main, dev, chore/add-proteina-cors-config]
     tags: ['v*.*.*']
 
 env:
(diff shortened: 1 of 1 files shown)


luislhl (approved):

luislhl (approved):

.github/workflows/deploy.yml
```diff
@@ -115,7 +115,7 @@ jobs:
           REDIS_PORT: 6379
           REDIS_DB: 0
           METADATA_BUCKET: hathor-explorer-metadata-testnet
-          CORS_ALLOWED_REGEX: https?://([a-z0-9]*\.){0,5}hathor\.network
+          CORS_ALLOWED_REGEX: https?:\/\/([a-z0-9]*\.){0,5}hathor\.network|https?:\/\/([a-z0-9\-]*\.?){0,5}hathor56\.wpcomstaging\.com
```
r4mmer, line 118: question: Can we have an example of the regex this is supposed to match?

From what i'm readind it can match up to 5 sub domains then `hathor56.wpcomstaging.com`.
So `https://foo.bar.hathor56.wpcomstaging.com` would be valid.

r4mmer (commented):
nitpick (non-blocking): I noted that the coverage is not passing, is this related to the network statistics issue?

.github/workflows/deploy.yml
alexruzenhack: You right! I left it open on purpose, but we can set a stricter value. We need to enable these two URLs:
- https://staging-94c0-hathor56.wpcomstaging.com
- https://hathor56.wpcomstaging.com/

alexruzenhack: Maybe we can use this regex: `https?:\/\/(staging-94c0-)?hathor56\.wpcomstaging\.com`. What do you think?

r4mmer: Great. didn't catch that you added an optional dot so we can use `...-hathor56...` 
But I think we can leave it open, make it easier to add more domains, as long as

r4mmer (approved):

PR #268, opened 2023-08-31
feat: network-statistics api

alexruzenhack:
### Acceptance Criteria
- Implement the `/network-statistics` endpoint
  - Uses elasticsearch to query transaction basic statistics such number of transactions, number of blocks and heighest height

### Security Checklist
- [ ] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #283, opened 2023-11-06
ci: tests for network statistics api

luislhl:
### Acceptance Criteria
- Add some unit tests for the network statistics API logic added in https://github.com/HathorNetwork/hathor-explorer-service/pull/268

### Motivation

We've been getting errors in the coverage check because we've dropped below 90%.

### Security Checklist

### DIFF
--- a/tests/unit/gateways/test_network_statistics_api_gateway.py
+++ b/tests/unit/gateways/test_network_statistics_api_gateway.py
@@ -1,13 +1,20 @@
 from unittest.mock import MagicMock
+
 from gateways.network_statistics_api_gateway import NetworkStatisticsApiGateway
+
 
 class TestNetworkStatisticsApiGateway:
     def setup_method(self):
         self.mock_elastic_search_client = MagicMock()
-        self.network_statistics_api_gateway = NetworkStatisticsApiGateway(self.mock_elastic_search_client)
+        self.network_statistics_api_gateway = NetworkStatisticsApiGateway(
+            self.mock_elastic_search_client
+        )
 
     def test_init(self):
-        assert self.network_statistics_api_gateway.elastic_search_client == self.mock_elastic_search_client
+        assert (
+            self.network_statistics_api_gateway.elastic_search_client
+            == self.mock_elastic_search_client
+        )
 
     def test_get_transaction_statistics(self):
         mock_response = {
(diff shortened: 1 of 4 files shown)


tests/unit/gateways/test_network_statistics_api_gateway.py
```diff
@@ -0,0 +1,36 @@
...
+    def setup_method(self):
+        self.mock_elastic_search_client = MagicMock()
+        self.network_statistics_api_gateway = NetworkStatisticsApiGateway(
+            self.mock_elastic_search_client
+        )
+
+    def test_init(self):
+        assert (
+            self.network_statistics_api_gateway.elastic_search_client
+            == self.mock_elastic_search_client
```
tuliomir, line 16: :star: Nice initial test, validating that the mocks are in place

tuliomir (approved):
Looks good to me.
Also: shouldn't we add the test coverage to the CI as well?

tests/unit/usecases/test_network_statistics_api.py
```diff
@@ -0,0 +1,43 @@
...
+        result = api.get_basic_statistics()
+        gateway.get_transaction_statistics.assert_called_once()
+        assert result == {"foo": "bar"}
+
+    @patch("usecases.network_statistics_api.logger")
+    def test_get_basic_statistics_gateway_error(self, logger, gateway):
+        api = NetworkStatisticsApi(network_statistics_api_gateway=gateway)
+        gateway.get_transaction_statistics.side_effect = exceptions.ApiError(
+            "gateway_error", MagicMock(), MagicMock()
+        )
```
alexruzenhack, line 28: 💬 I would suggest here to not use the "gate_error" as message but another text, and use the "gateway_error" to assert the error raised in NetworkStatisticsApi. This one should contain the "gate_error" message. What do you think?

If you agree, it should be applied in the next test as well.

alexruzenhack (approved):

tests/unit/usecases/test_network_statistics_api.py
luislhl: Done in https://github.com/HathorNetwork/hathor-explorer-service/pull/283/commits/b0d8d1a1916916621bbca8b20e94494f4d3649c5

I had to change the structure a little bit to be able to assert the message of the raised Exception.

luislhl:
> Also: shouldn't we add the test coverage to the PR context as well? Like what we have with the wallets repositories and the _Codecov_ tool.

@tuliomir , I think that's a good idea, but I think it fits better in a new PR. This one was just to fix the "emergency" of having the CI failing because of the coverage.

Adding Codecov would be something else.

PR #290, opened 2023-12-18
fix: use latest python-healthchecklib

luislhl:
### Acceptance Criteria
- The new version of python-healthchecklib fixes a bug that was affecting us. More info in https://github.com/HathorNetwork/python-healthcheck-lib/pull/2

### Security Checklist

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1157,13 +1157,13 @@ cli = ["click (>=5.0)"]
 
 [[package]]
 name = "python-healthchecklib"
-version = "0.1.0"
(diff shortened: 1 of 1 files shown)


r4mmer (approved):
question(non-blocking): Do we have a list of downstream projects for python-healthchecklib?

If not, I think we should create it so that whenever we have an update on the lib we can check the projects that use it and update them too (like we are doing in this PR, but more proactively).

andreabadesso (approved):

PR #322, opened 2024-04-18
fix: function name is bigger than 64 chars

pedroferreira1:
### Motivation

The lambda deploy failed because the lambda function name after adding the prefix of the app and staging was bigger than 64 characters (`hathor-explorer-service-mainnet-node_api_nc_blueprint_information` has 65 chars).

### Acceptance Criteria
- Function name (`node_api_nc_blueprint_information`) must have 1 less char.

### Security Checklist

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -693,7 +693,7 @@ functions:
               - name: request.querystring.count
               - name: request.header.origin
 
-  node_api_nc_blueprint_information:
+  node_api_nc_blueprint_info:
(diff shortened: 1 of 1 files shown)


r4mmer (approved):
I think we should change the `hathor-explorer-service-mainnet-` prefix as well in the future.
To something like `explorer-svc-mainnet-` to make this limit harder to hit in the future.

tuliomir (approved):

PR #330, opened 2024-07-15
ci: deploy to alternative testnets

luislhl:
### Acceptance Criteria
- Make sure we deploy to our alternative testnets every time we deploy to the main testnet

### Security Checklist


alexruzenhack (approved):

.github/workflows/deploy.yml
```diff
@@ -156,6 +155,111 @@ jobs:
           export DOCKER_IMAGE_TAG=testnet-${{ github.sha }}-$timestamp
           make deploy-daemons
           rm /home/runner/.docker/config.json
+  deploy-nano-testnet:
```
r4mmer, line 158: suggestion: Maybe change this into a reusable workflow.

The [github reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) allows us to create workflows in a separate file and just invoke it with some parameters, which would be less repetitive and easier to update in the future since updates would affect all deploy jobs.

.github/workflows/deploy.yml
```diff
@@ -156,6 +155,111 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [init, dependencies]
+
+    if: ${{ needs.init.outputs.environment == 'testnet' }}
+
+    steps:
+        # https://github.com/actions/checkout/releases/tag/v3.5.3
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9
+      - name: Prepare deploy
+        uses: ./.github/workflows/lib/prepare-deploy
```
r4mmer, line 168: question(blocking): Where is this defined?

Could not find this file.

.github/workflows/deploy.yml
luislhl: I did this in the base PR: https://github.com/HathorNetwork/hathor-explorer-service/pull/329

It seemed better indeed

r4mmer (approved):

alexruzenhack (approved):

PR #331, opened 2024-08-16
feat: add before parameter to the nc history API

pedroferreira1:
### Acceptance Criteria
- Add `before` parameter to the nc history API

### Security Checklist


r4mmer (approved):
suggestion(non-blocking): Add tests for nc apis.

I know this repo does not have coverage enabled but we should add these new apis.
Maybe we should do this on another PR, waht do you think?

tuliomir (approved):

PR #332, opened 2024-08-19
fix: hathor-core healthcheck

luislhl:
### Acceptance Criteria
- We should expect to receive a response with Content-Type `text/html` from hathor-core's health endpoint, and make aiohttp parse it without errors.
- After https://github.com/HathorNetwork/hathor-core/pull/1110 gets released, we will be able to revert this change. I preferred to do it instead of waiting for hathor-core, because it could take too long.

### Security Checklist

### DIFF
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -43,7 +43,11 @@ class HathorCoreAsyncClient:
         self.log = logger.new(client="async")
 
     async def get(
-        self, path: str, params: Optional[dict] = None, timeout: Optional[float] = None
+        self,
+        path: str,
+        params: Optional[dict] = None,
+        timeout: Optional[float] = None,
+        content_type: Optional[str] = "application/json",
     ) -> Dict[Any, Any]:
         """Make a get request async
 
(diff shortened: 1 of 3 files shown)


alexruzenhack (approved):

gateways/clients/hathor_core_client.py
```diff
@@ -71,7 +75,7 @@ async def get(
                             status=response.status,
                             body=await response.text(),
                         )
-                    return await response.json()
+                    return await response.json(content_type=content_type)
```
r4mmer, line 78: question(blocking): What is the purpose of this `content_type`?

luislhl: It acts as the expected Content-Type for aiohttp. It will check whether the response Content-Type matches the expected one, and raise an error if not.

Their default is `application/json`, so they reject parsing as JSON anything that has a different Content-Type.

My change will allow us to avoid aiohttp raising an error when hathor-core returns a `text/html` from the `/health` endpoint.

I will pass `None` to it, which basically disables this check. I'm doing this instead of passing `text/html` to make sure we won't break the explorer-service after https://github.com/HathorNetwork/hathor-core/pull/1110 gets deployed

r4mmer (approved):

PR #333, opened 2024-08-19
feat: add new hathor core api to get blueprint source code

pedroferreira1:
### Acceptance Criteria
- Add hathor-core API to get blueprint source code.

### Security Checklist


r4mmer (approved):

handlers/node_api.py
```diff
@@ -392,3 +392,28 @@ def nc_blueprint_information(
...
+    blueprint_id = event.query.get("blueprint_id")
+
+    if blueprint_id is None:
+        raise ApiError("invalid_parameters")
+
+    # This might throw HathorCoreTimeout error
+    response = node_api.get_nc_blueprint_source_code(blueprint_id)
+
+    if response is None or "error" in response:
+        message = response.get("error") if (response and "error" in response) else ""
```
tuliomir, line 412: question(non-blocking): Should we add a default error message here? It may help with eventual debuggings

```suggestion
        message = response.get("error") if (response and "error" in response) else "Unknown error when retrieving blueprint source code"
```

serverless.yml
```diff
@@ -714,6 +714,26 @@ functions:
...
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: node_api/nc_blueprint_source_code
+          method: get
+          cors: true
+          caching:
+            enabled: true
+            ttlInSeconds: 300
```
tuliomir, line 732: question(non-blocking): Is there any downside to storing this for longer periods of time? Days, maybe months? As far as I know, blueprint source codes are immutable, except for new breaking versions of the fullnode.

tuliomir (approved):

serverless.yml
pedroferreira1: The source code might change as long as it maintains compatibility. We could improve a docstring, add optional metadata, etc.

handlers/node_api.py
pedroferreira1: Done https://github.com/HathorNetwork/hathor-explorer-service/pull/333/commits/39a0278b95df3e7bea194618910b917022fae587

PR #350, opened 2024-10-23
feat: truncate children in tx detail API to return at most 100 elements

pedroferreira1:
### Motivation

There are networks with low quantity of transactions that remain for a long time without transactions. Because of that, the latest tx keeps being used by blocks as parent, so it list of children increases a lot. The explorer service lambda returns 502 for any response bigger than 6mb, so some transactions (https://explorer.testnet.hathor.network/transaction/00e161a6b0bee1781ea9300680913fb76fd0fac4acab527cd9626cc1514abdc9, for example) were not being loaded.

There's no reason for the UI to receive a list of 193k elements. We should paginate it, if we decide that it's important to show all of them. For now, we are truncating it and returning the latest 100 elements.

### Acceptance Criteria
- Truncate children list in the tx detail API to return at most 100 elements.

### Security Checklist

### DIFF
--- a/handlers/node_api.py
+++ b/handlers/node_api.py
@@ -176,6 +176,12 @@ def get_transaction(
     if id is None:
         raise ApiError("invalid_parameters")
     response = node_api.get_transaction(id)
+    # It does not make sense to show in the explorer thousands of children.
+    # The full node will continue returning the correct data but in the explorer
+    # service we will truncate it and return only the latest 100 to show in the UI
+    # (it might even be more than we should, maybe we should paginate in the future)
+    if 'meta' in response and 'children' in response['meta']:
+        response['meta']['children'] = response['meta']['children'][-100:]
 
     return {
         "statusCode": 200,
--- a/handlers/node_api.py
+++ b/handlers/node_api.py
@@ -180,8 +180,8 @@ def get_transaction(
     # The full node will continue returning the correct data but in the explorer
     # service we will truncate it and return only the latest 100 to show in the UI
     # (it might even be more than we should, maybe we should paginate in the future)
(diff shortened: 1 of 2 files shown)


luislhl (approved):

handlers/node_api.py
```diff
@@ -176,6 +176,12 @@ def get_transaction(
     if id is None:
         raise ApiError("invalid_parameters")
     response = node_api.get_transaction(id)
+    # It does not make sense to show in the explorer thousands of children.
+    # The full node will continue returning the correct data but in the explorer
+    # service we will truncate it and return only the latest 100 to show in the UI
+    # (it might even be more than we should, maybe we should paginate in the future)
+    if 'meta' in response and 'children' in response['meta']:
+        response['meta']['children'] = response['meta']['children'][-100:]
```
r4mmer, line 184: suggestion: Move 100 to the config.

```suggestion
        response['meta']['children'] = response['meta']['children'][-TX_DAG_CHILDREN_CUTOFF:]
```
Or `MAX_TX_CHILDREN`.

r4mmer, line 184: Personally I think we can be stricter with this.
We can just cap at 20 instead.

pedroferreira1: https://github.com/HathorNetwork/hathor-explorer-service/pull/350/commits/7dd79d05923726c3898b36043b2670ca15e91e66 done

pedroferreira1: I agree with you, 20 should be more than enough. But we are currently showing everything, so I prefer keeping it as a "big" number for now, and we can reduce it in the future if we prefer.

r4mmer (approved):

PR #396, opened 2026-04-20
Add metadata/icon endpoint (S3 icons) and binary PNG support

luislhl:
What
- Add a new endpoint `GET /metadata/icon?id=<id>` that proxies PNG icons stored in the metadata S3 bucket under `icons/{id}.png`.

Changes
- gateways/clients/s3_client.py: add `load_file_bytes()` to fetch raw bytes from S3.
- gateways/metadata_gateway.py: add `get_icon_metadata(id)` to read `icons/{id}.png`.
- handlers/icon_metadata_handler.py: new Lambda that returns base64 body with `isBase64Encoded: true` and `Content-Type: image/png`.
- usecases/metadata.py: register the `icon` type so it can be fetched via the usecase layer.
- serverless.yml: add IAM `s3:GetObject` permission for `icons/*`, add `get_icon_metadata_handler` function, and configure `binaryMediaTypes: ['image/png']` with a note on `Accept` header tradeoffs.
- tests: add unit tests to cover S3 bytes loading and metadata icon flow.

Why / Decisions
- API Gateway requires binary media types + correct `Accept` header to decode base64 responses to raw bytes. Browsers typically send the appropriate Accept header, but CLI clients (wget/curl) may not — they must send `Accept: image/png` or you can keep `'*/*'` in `binaryMediaTypes` (tradeoff documented in serverless.yml).
- Lambdas return base64 + `isBase64Encoded: true`; API Gateway performs decoding when request headers/content-type match the binary media types.

How to test
1) We deployed this to our test environment:
   wget --header="Accept: image/png" \
  "https://dev.explorer-service.testnet.hathor.network/metadata/icon?id=0000000073dd17b4c0f4670da70b7a2c9a07a629e38b9a47196c0fe7335f333e" -O 0000000073dd17b4c0f4670da70b7a2c9a07a629e38b9a47196c0fe7335f333e.png

## Summary by CodeRabbit

* **New Features**
  * Added icon metadata retrieval via GET /metadata/icon.
  * API now supports PNG/binary responses so images are served correctly.

* **Performance**
  * Endpoint caching enabled (TTL 86400s) and varies by id, origin, and Accept header.

* **Tests**
  * Added unit tests covering S3 byte retrieval, metadata gateway, and use-case integration for icons.

### DIFF
--- a/serverless.yml
+++ b/serverless.yml
@@ -424,7 +424,7 @@ functions:
           cors: true
           caching:
             enabled: true
-            ttlInSeconds: 300
+            ttlInSeconds: 86400  # one day cache, as icons don't change often
             cacheKeyParameters:
               - name: request.querystring.id
               - name: request.header.origin
--- a/serverless.yml
+++ b/serverless.yml
@@ -20,6 +20,16 @@ provider:
   apiGateway:
     apiKeys:
       - explorer-service-${self:custom.stage}-healthcheck-api-key
+    # API Gateway decodes Lambda responses marked as base64 only when the
+    # request's `Accept` header matches one of the configured binary media types.
+    #
+    # Clients (wget, curl) should send `Accept: image/png` to receive raw PNG
+    # bytes. We could include '*/*' here for broader compatibility so generic clients
+    # would also receive binary responses, but that could affect how API Gateway treats
+    # other responses if Content-Type headers are missing. We preferred letting the clients
+    # set the client `Accept` header.
+    binaryMediaTypes:
+      - 'image/png'
   logs:
     restApi: # TODO: add data to make the serverless create this role by itself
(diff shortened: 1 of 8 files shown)


r4mmer (approved):

serverless.yml
```diff
@@ -393,6 +409,27 @@ functions:
...
+    layers:
+      - { Ref: PythonRequirementsLambdaLayer }
+    events:
+      - http:
+          path: metadata/icon
+          method: get
+          cors: true
+          caching:
+            enabled: true
+            ttlInSeconds: 300
```
pedroferreira1, line 427: It seems to me that the icons won't change frequently. Maybe we could increase this `ttl`. If it's easy to delete this cache every time we add a new icon or update an existing one, this could be even a day

pedroferreira1 (approved):

serverless.yml
luislhl: Done. Changed to 1 day.

PR #397, opened 2026-04-24
feat: add retry logic for 5xx gateway errors in async HTTP client

luislhl:
## Problem

When a fullnode is temporarily unreachable (e.g. nginx returns `502 Bad Gateway`), the daemon was generating **three log entries per node per 1-second polling cycle**, triggering spurious alerts:

```
1. WARNING  hathor_core_error      {client: async, path: /v1a/status, status: 502, body: <html>...}
2. ERROR    hathor_core_error      {client: async, path: /v1a/status, error: ContentTypeError(...)}
3. WARNING  collect_status_error   {error: ContentTypeError(...)}
```

### Root cause

`HathorCoreAsyncClient.get()` had no retry mechanism, and `response.json()` was called unconditionally:

1. Any non-2xx status code immediately logged a `WARNING` with the raw body.
2. `response.json()` was then called on a 502 response whose body is nginx HTML — this raised a `ContentTypeError`, caught by the `except` branch, which logged an `ERROR` and returned `{"error": repr(exc)}` (the original body lost).
3. `_send()` in `CollectNodesStatuses` saw the `"error"` key in the returned dict and logged a third `WARNING`.

### 1. Retry transient gateway errors

`HathorCoreAsyncClient.get()` now retries retryable gateway errors (`502`, `503`, `504`):

- **`MAX_RETRIES = 2`** — up to 2 retries before giving up
- **`RETRY_DELAY = 1.0s`** — wait 1 second between attempts
- **`RETRYABLE_STATUS_CODES = {502, 503, 504}`** — only gateway-level transient errors are retried

All attempts share a single `aiohttp.ClientSession`. On a retryable status the body is read (consuming the connection cleanly) and the loop either sleeps + retries or, once retries are exhausted, logs a single `WARNING` and returns the decoded body (see below). Non-retryable 4xx errors are never retried. The `get()` docstring now also clarifies that `timeout` is **per-attempt**, so total elapsed time may exceed it when retries happen.

### 2. Decode error bodies instead of unconditionally calling `response.json()`

For any error response (`status > 299`), `_decode_error_body` returns:

- the server-provided **JSON body as-is** when the body is valid JSON (callers keep the full error details — this matches `main`'s behavior for JSON error bodies);
- `{"error": "status X", "body": <raw body>}` when the body is **not** JSON (e.g. GCP load balancer / nginx 5xx that come back as HTML/plain text).

This is the key fix versus `main`: a non-JSON error body no longer reaches `response.json()`, so it no longer raises a `ContentTypeError` (eliminating the spurious `ERROR` log) and the raw error text is preserved instead of being replaced by `repr(exc)`. Successful (`<= 299`) responses are unchanged: the JSON body is returned verbatim.

### 3. Retry opt-out for the health endpoint

Because a `503` from `/v1a/health` is a *legitimate answer* (unhealthy fullnode) rather than a transient gateway failure, retrying it would only burn the tight healthcheck timeout budget (5s client / 6s lambda). `get()` takes a `retry` flag (default `True`) and `HealthcheckGateway` passes `retry=False`, so genuine gateway errors are still retried everywhere else.

### Outcome (per failing poll cycle, vs `main`)

| Scenario | `main` | This PR |
|---|---|---|
| Transient 502 (recovers within retries) | 1 WARNING + 1 ERROR + 1 WARNING = 3 | **0** |
| Persistent 502 (all retries fail) | 1 WARNING + 1 ERROR + 1 WARNING = 3 | **2** — `hathor_core_error` WARNING + `collect_status_error` WARNING; the spurious `ERROR` is gone |

## Why this fits the two current callers

This client is consumed in only two places, and neither relies on the HTTP status code to detect errors — which matches hathor-core's convention of **frequently returning errors with HTTP 200 and signalling them in the body** (many resources `return {'success': False, ...}` without ever setting a response code). Error detection here is intentionally body-based, not status-based.

| Caller | Endpoint | hathor-core behavior | How the caller reads the result |
|---|---|---|---|
| `CollectNodesStatuses` (`usecases/collect_nodes_statuses.py`) | `/v1a/status` | `StatusResource.render_GET` **always returns HTTP 200** with the status dump — no error branch, never sets a response code. The only failures observable here are transport-level (timeout / connection refused) → `{"error": repr(e)}`, plus upstream gateway 5xx from nginx/LB. | Checks `if "error" in data`, otherwise parses the body with `Node.from_status_dict`. Since the endpoint itself never returns an error status, it only reaches `_decode_error_body` for upstream gateway errors (non-JSON → `{"error": ..., "body": ...}`, still caught by the `"error"` check). |
| `GetHealthcheck` (`usecases/get_healthcheck.py` via `HealthcheckGateway`) | `/v1a/health` | `HealthcheckResource` returns **200 when healthy and 503 when unhealthy**, but the verdict is always in the body (`status: pass/fail`). It also accepts `strict_status_code=1` to force 200 even when failing. | Reads `health_response["status"]` for the `pass`/`warn`/`fail` nuance, so it needs the 503 body returned as-is — which `_decode_error_body` does (the body is a JSON object). Retries are disabled here so the legitimate 503 is returned immediately. |

## Changes

- `gateways/clients/hathor_core_client.py` — retry loop in `HathorCoreAsyncClient.get()`; `MAX_RETRIES` / `RETRY_DELAY` / `RETRYABLE_STATUS_CODES` constants; `_decode_error_body`; `retry` opt-out flag; per-attempt `timeout` docstring
- `gateways/healthcheck_gateway.py` — pass `retry=False` for the health endpoint
- `tests/unit/gateways/clients/test_hathor_core_client.py` — new `TestHathorCoreAsyncClientGet` covering success, retry-then-success, single session reuse, retry exhaustion, non-retryable 4xx, JSON/non-JSON error bodies, the `retry=False` opt-out, and network exceptions
- `tests/unit/gateways/test_healthcheck_gateway.py` — assert `retry=False` is passed for the health call

## Summary by CodeRabbit

* **New Features**
  * Added transient retry support for gateway requests via an opt-in retry mode, with per-attempt timeout handling.
* **Bug Fixes**
  * Healthcheck requests now explicitly avoid retries to preserve the healthcheck timeout budget.
  * Improved consistent error normalization for non-JSON/JSON failure responses, including behavior when retries are exhausted.
* **Tests**
  * Expanded unit coverage for retry success, retry exhaustion, non-retryable 4xx behavior, non-JSON vs JSON error handling, and network exception scenarios.

### DIFF
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -101,6 +101,8 @@ class HathorCoreAsyncClient:
             self.log.error("hathor_core_error", path=path, error=repr(e))
             return {"error": repr(e)}
 
+        return {"error": "max retries exceeded"}
+
     async def post(
         self, path: str, body: Optional[dict] = None, timeout: Optional[float] = None
     ) -> Dict[Any, Any]:
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -81,7 +81,9 @@ class HathorCoreAsyncClient:
                     async with session.get(url, params=params) as response:
                         if response.status > 299:
                             body = await response.text()
-                            is_retryable = response.status in self.RETRYABLE_STATUS_CODES
+                            is_retryable = (
+                                response.status in self.RETRYABLE_STATUS_CODES
+                            )
 
                             if is_retryable and attempt < self.MAX_RETRIES:
                                 await asyncio.sleep(self.RETRY_DELAY)
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -81,11 +81,9 @@ class HathorCoreAsyncClient:
                     async with session.get(url, params=params) as response:
                         if response.status > 299:
                             body = await response.text()
+                            is_retryable = response.status in self.RETRYABLE_STATUS_CODES
 
-                            if (
-                                response.status in self.RETRYABLE_STATUS_CODES
-                                and attempt < self.MAX_RETRIES
-                            ):
+                            if is_retryable and attempt < self.MAX_RETRIES:
                                 await asyncio.sleep(self.RETRY_DELAY)
                                 continue
                             self.log.warning(
@@ -94,8 +92,9 @@ class HathorCoreAsyncClient:
                                 status=response.status,
                                 body=body,
                             )
-
-                            return {"error": f"status {response.status}"}
+                            if is_retryable:
+                                return {"error": f"status {response.status}"}
+                            return await response.json(content_type=content_type)
                         return await response.json(content_type=content_type)
         except Exception as e:
             self.log.error("hathor_core_error", path=path, error=repr(e))
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -64,7 +64,8 @@ class HathorCoreAsyncClient:
         :type path: str
         :param params: params to be sent
         :type params: Optional[dict]
-        :param timeout: timeout in seconds
+        :param timeout: per-attempt timeout in seconds; total elapsed time may be higher
+            when retries are triggered
         :type timeout: Optional[float]
         """
         url = parse.urljoin(self.url, path)
@@ -72,15 +73,19 @@ class HathorCoreAsyncClient:
         if not timeout:
             timeout = self.DEFAULT_TIMEOUT
 
-        for attempt in range(self.MAX_RETRIES + 1):
-            try:
-                async with aiohttp.ClientSession(
-                    timeout=aiohttp.ClientTimeout(total=timeout)
-                ) as session:
+        try:
+            async with aiohttp.ClientSession(
+                timeout=aiohttp.ClientTimeout(total=timeout)
+            ) as session:
+                for attempt in range(self.MAX_RETRIES + 1):
                     async with session.get(url, params=params) as response:
-                        if response.status in self.RETRYABLE_STATUS_CODES:
+                        if response.status > 299:
                             body = await response.text()
-                            if attempt < self.MAX_RETRIES:
+
+                            if (
+                                response.status in self.RETRYABLE_STATUS_CODES
+                                and attempt < self.MAX_RETRIES
+                            ):
                                 await asyncio.sleep(self.RETRY_DELAY)
                                 continue
                             self.log.warning(
@@ -89,18 +94,12 @@ class HathorCoreAsyncClient:
                                 status=response.status,
                                 body=body,
                             )
+
                             return {"error": f"status {response.status}"}
-                        if response.status > 299:
-                            self.log.warning(
-                                "hathor_core_error",
-                                path=path,
-                                status=response.status,
-                                body=await response.text(),
-                            )
                         return await response.json(content_type=content_type)
-            except Exception as e:
-                self.log.error("hathor_core_error", path=path, error=repr(e))
-                return {"error": repr(e)}
+        except Exception as e:
+            self.log.error("hathor_core_error", path=path, error=repr(e))
+            return {"error": repr(e)}
 
     async def post(
         self, path: str, body: Optional[dict] = None, timeout: Optional[float] = None
--- a/gateways/clients/hathor_core_client.py
+++ b/gateways/clients/hathor_core_client.py
@@ -1,3 +1,4 @@
+import asyncio
 import json
 from typing import Any, Dict, Optional
 from urllib import parse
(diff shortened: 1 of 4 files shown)


gateways/clients/hathor_core_client.py
```diff
@@ -72,19 +77,33 @@ async def get(
...
+                            )
+                            if is_retryable:
+                                return {"error": f"status {response.status}"}
+                            return await response.json(content_type=content_type)
+                        return await response.json(content_type=content_type)
         except Exception as e:
             self.log.error("hathor_core_error", path=path, error=repr(e))
             return {"error": repr(e)}
 
+        return {"error": "max retries exceeded"}
```
luislhl, line 110: Improve this after Carneiro conversation:

- We could add more info about what happened exactly (like what errors we got) for the retries to fail.

gateways/clients/hathor_core_client.py
```diff
@@ -72,19 +77,33 @@ async def get(
...
+                                await asyncio.sleep(self.RETRY_DELAY)
+                                continue
+                            self.log.warning(
+                                "hathor_core_error",
+                                path=path,
+                                status=response.status,
+                                body=body,
+                            )
+                            if is_retryable:
+                                return {"error": f"status {response.status}"}
```
luislhl: Improve this after Carneiro conversation:

- We should try to check whether the response is really not JSON
- We should improve the error message if it's not, like with the whole body which should contain some error.

luislhl: Done in 43e11e8. Following the conversation with @r4mmer, instead of overriding error responses with a synthetic `{"error": "status X"}`, we now check whether the body is really JSON before parsing it: if it is, we return the server-provided JSON as-is; if it isn't (e.g. GCP LB 5xx coming back as HTML/text), we return `{"error": "status X", "body": <raw body>}` so the actual error text is preserved. This goes through the new `_decode_error_body` helper and applies to both non-retryable and retryable-after-exhaustion paths.

gateways/clients/hathor_core_client.py
luislhl: Addressed in 43e11e8. The retryable-after-exhaustion path now returns the actual last response body via `_decode_error_body` (the raw error text from upstream), so the failure context is preserved instead of a generic message. The `return {"error": "max retries exceeded"}` line is now effectively an unreachable safety net (the loop always returns on its final attempt), so I left it as-is.

gateways/clients/hathor_core_client.py
```diff
@@ -72,15 +77,26 @@ async def get(
...
+                                continue
+                            self.log.warning(
+                                "hathor_core_error",
+                                path=path,
+                                status=response.status,
+                                body=body,
+                            )
+
+                            return {"error": f"status {response.status}"}
+                        return await response.json(content_type=content_type)
```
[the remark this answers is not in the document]
luislhl: Resolved. Non-retryable responses no longer get the synthetic `{"error": "status X"}`: in 43e11e8 we return the server-provided JSON body when present (and the status + raw body otherwise) via `_decode_error_body`, for both 4xx and non-retryable 5xx.

gateways/clients/hathor_core_client.py
```diff
@@ -68,22 +72,35 @@ async def get(
...
-                    return await response.json(content_type=content_type)
-        except Exception as e:
-            self.log.error("hathor_core_error", path=path, error=repr(e))
-            return {"error": repr(e)}
+        for attempt in range(self.MAX_RETRIES + 1):
+            try:
+                async with aiohttp.ClientSession(
+                    timeout=aiohttp.ClientTimeout(total=timeout)
+                ) as session:
+                    async with session.get(url, params=params) as response:
```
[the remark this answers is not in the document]
luislhl: Addressed earlier in 4ca3310 — a single `aiohttp.ClientSession` is now created outside the retry loop and reused across attempts.

gateways/clients/hathor_core_client.py
```diff
@@ -68,22 +72,35 @@ async def get(
...
+                        if response.status in self.RETRYABLE_STATUS_CODES:
+                            body = await response.text()
+                            if attempt < self.MAX_RETRIES:
+                                await asyncio.sleep(self.RETRY_DELAY)
```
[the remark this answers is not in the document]
luislhl: Intentional for now: `timeout` is per-attempt (documented in the `get` docstring), so total elapsed time can exceed it when retries fire. The daemon/healthcheck callers were the motivation for the retries in the first place, and we'd rather retry than fail fast on a transient 5xx. We may add a separate overall-deadline parameter later if a caller needs a hard budget.

r4mmer (approved):

raul-oliveira (approved):

